PR #12, opened 2021-06-06
Fix typo in docs

a8t:
Thought it was a strange API, and then wondered why the strange API didn't work. Turns out it was just a small typo in the docs 😄

Thanks for this library!


wojtekmach:
Thanks!

PR #35, opened 2021-08-12
Support plugins

wojtekmach:
Demo:

image

See:

- https://github.com/wojtekmach/notebooks/blob/main/req_plugins.livemd
- https://github.com/wojtekmach/req_s3/blob/main/lib/req_s3.ex
- https://github.com/wojtekmach/req_hex/blob/main/lib/req_hex.ex


wojtekmach:
Closing while we wait for #44.

PR #37, opened 2021-09-06
Update readme (request.uri)

thbar:
I believe the field name is now `uri` (per https://github.com/wojtekmach/req/blob/b397a0aad1e8314350b222aa04a81cc4e1e2f24b/lib/req/request.ex#L9).


wojtekmach:
Thank you!

PR #39, opened 2021-09-12
Ensure get_private/put_private have atom keys

### LINKED ISSUE #28: get_private, put_private: Ensure keys are atoms
Let's do this now so that in the future we might support "paths", .eg.: `put_private(request, [:foo, :bar], :baz)`.

mhanberg:
Closes #28


wojtekmach:
Thank you!

PR #51, opened 2021-11-09
Add `:private` field to Response

heydtn:
It was possible to annotate Requests with library-specific metadata using the :private field, however the same wasn't true for Responses. This means that it wasn't possible to return extra information to the caller.

### DIFF
--- a/lib/req/response.ex
+++ b/lib/req/response.ex
@@ -9,7 +9,30 @@ defmodule Req.Response do
     * `:headers` - the HTTP response headers
 
     * `:body` - the HTTP response body
+
+    * `:private` - a map reserved for libraries and frameworks to use.
+      Prefix the keys with the name of your project to avoid any future
+      conflicts. Only accepts `t:atom/0` keys.
   """
 
-  defstruct [:status, headers: [], body: ""]
+  defstruct [
+    :status,
+    headers: [],
+    body: "",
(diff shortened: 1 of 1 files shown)


wojtekmach:
Thank you for the PR, I think it is a good addition. Before going forward with it, could you give a couple examples what you'd use this for?

heydtn:
Of course!  Off-hand, I can think of a few.

1. If a library wants to keep more information about retry attempts, it may want to record why each retry happened, or how long each retry took, etc.
2. If a library wants to keep a list of all redirects followed while running a request.
3. Returning timestamps for when the request was sent and when the response was obtained. This could theoretically be a separate/core feature, but I thought I'd mention it anyway.
4. Normalization and record keeping. For example, HTTP says that headers are case-insensitive, however there are servers that treat headers as case-sensitive. It would be possible to have a Request step which normalizes Request headers into a format the server expects, but then it's not possible to return the normalized request to the caller so it knows what was actually sent.

There's other potential features that I can't really go into detail about, though most revolve around tracking in-depth details, annotating, and returning more information from the request/response lifecycle.  They're niche and don't particularly make sense for inclusion in the standard library, which is why `:private` makes the most sense.

wojtekmach:
Thank you, these are all very compelling use cases.

PR #54, opened 2021-11-11
Update Finch to 0.9.1

QuinnWilton:
Hey Wojtek!

We're relying on the latest version of `Finch`, so I just wanted to bump the version in `Req` so that we could remove an `override: true` in our `mix.exs`. Everything should be backwards compatible, and I think `Req` actually already relied on `~> 0.9.0` because that's when support for Unix sockets was added.


wojtekmach:
Thx, Ill cut a release tomorrow

QuinnWilton:
Thank you!

coop:
Thanks @QuinnWilton - I was just about to open a PR. @wojtekmach any chance you'd be able to cut a release? :)

wojtekmach:
v0.2.1 is out, sorry for the hold up!

coop:
Thank you so much @wojtekmach - also thanks for the lib, it's cut down on the boilerplate I used to write.

Issue #64, opened 2022-01-06
`follow_redirects`: Add `:max_redirects` option


wojtekmach:

aleDsz:
Should it have a `:redirect_delay` like retry does?

wojtekmach:
I don't think so, why would we want to wait?

aleDsz:
Maybe the host that we are making the request could have some protection against multiple requests in the same time

wojtekmach:
I see, fair enough. Let's not worry about this at the moment, can always address it later.

PR #67, opened 2022-01-17
`follow_redirects`: Don't send auth headers on redirect to different host

### LINKED ISSUE #13: `follow_redirects`: Do not send authorization header on redirects to different hosts
See `curl --locattion` and `curl --location-trusted` for more information.

srowley:
Closes #13.

### DIFF
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -24,6 +24,9 @@ defmodule Req.Request do
 
     * `:error_steps` - the list of error steps
 
+    * `:location_trusted` - if `false`, authorization headers are not
+      sent to different hosts on redirects. Defaults to `false`.
+
     * `:private` - a map reserved for libraries and frameworks to use.
       Prefix the keys with the name of your project to avoid any future
       conflicts. Only accepts `t:atom/0` keys.
@@ -124,7 +127,8 @@ defmodule Req.Request do
     request_steps: [],
     response_steps: [],
     error_steps: [],
-    private: %{}
+    private: %{},
+    location_trusted: false
   ]
 
   @doc """
@@ -186,7 +190,8 @@ defmodule Req.Request do
       private: %{
         req_finch:
           {Keyword.get(options, :finch, Req.Finch), Keyword.get(options, :finch_options, [])}
-      }
+      },
+      location_trusted: Keyword.get(options, :location_trusted, false)
     }
   end
 
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -24,9 +24,6 @@ defmodule Req.Request do
 
     * `:error_steps` - the list of error steps
 
-    * `:location_trusted` - if `false`, authorization headers are not
-      sent to different hosts on redirects. Defaults to `false`.
-
     * `:private` - a map reserved for libraries and frameworks to use.
       Prefix the keys with the name of your project to avoid any future
       conflicts. Only accepts `t:atom/0` keys.
@@ -190,8 +187,7 @@ defmodule Req.Request do
       private: %{
         req_finch:
           {Keyword.get(options, :finch, Req.Finch), Keyword.get(options, :finch_options, [])}
-      },
-      location_trusted: Keyword.get(options, :location_trusted, false)
+      }
     }
   end
 
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -36,7 +36,7 @@ defmodule Req.Steps do
     * [`&retry(&1, options[:retry])`](`retry/2`) (if `options[:retry]` is set to
       an atom true or a options keywords list)
 
-    * `follow_redirects/1`
+    * [`&follow_redirects(&1, options[:follow_redirects])`](`follow_redirects/2`)
 
     * `decompress/1`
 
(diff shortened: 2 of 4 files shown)


lib/req/request.ex
```diff
@@ -24,6 +24,9 @@ defmodule Req.Request do
 
     * `:error_steps` - the list of error steps
 
+    * `:location_trusted` - if `false`, authorization headers are not
+      sent to different hosts on redirects. Defaults to `false`.
```
wojtekmach, line 28: In the future we also want to control redirects with `:max_redirects` option (https://github.com/wojtekmach/req/issues/64) so instead of adding a new field to the struct, let's change `follow_redirects/1` to `follow_redirects/2` where the 2nd argument is options keyword list.

See how `retry` is configured in `put_default_steps` for a similar idea.

lib/req/steps.ex
```diff
@@ -829,4 +825,31 @@ defmodule Req.Steps do
...
+    else
+      url = URI.parse(location)
+      put_in(request.url, url)
+    end
+  end
+
+  defp remove_credentials_if_untrusted(%{location_trusted: true} = request, _), do: request
+
+  defp remove_credentials_if_untrusted(%{url: url} = request, location) do
+    if URI.parse(location).host == url.host do
```
wojtekmach, line 842: it seems we're calling `URI.parse(location)` here for the 2nd time, first time was in `put_redirect_location`. Could you change it to do the parsing just once? Maybe like this?

```elixir
  location = URI.parse(location)
  request =
    request
    |> remove_credentials_if_untrusted(location)
    |> put_redirect_location(location)
```

lib/req/steps.ex
```diff
@@ -829,4 +825,31 @@ defmodule Req.Steps do
...
+    else
+      remove_credentials(request)
+    end
+  end
+
+  defp remove_credentials(request) do
+    headers = List.keydelete(request.headers, "authorization", 0)
+    request_steps = Enum.reject(request.request_steps, fn {_, step, _} -> step == :auth end)
+
+    struct(request, headers: headers, request_steps: request_steps)
```
wojtekmach, line 853:
```suggestion
    %{request | headers: headers, request_steps: request_steps}
```

test/req/steps_test.exs
```diff
@@ -353,6 +353,88 @@ defmodule Req.StepsTest do
...
+      |> Plug.Conn.send_resp(302, "redirecting to #{location}")
+    end)
+
+    Bypass.expect(c.bypass, "GET", "/auth", fn conn ->
+      assert auth_header in conn.req_headers
+      Plug.Conn.send_resp(conn, 200, "ok")
+    end)
+
+    assert ExUnit.CaptureLog.capture_log(fn ->
+             assert Req.get!(c.url <> "/redirect", auth: {"foo", "bar"}).status == 200
```
wojtekmach, line 398: I think we're missing setting `location_trusted: true`

wojtekmach (commented):
This is great, I left a few comments below.

srowley:
Thanks - these all make sense. My GitHub collaboration knowledge is also at
the relative novice stage - should I squash these changes into this commit
and then force push, or something else?

Steve

wojtekmach:
You don't have to manually squash as I can easily do that when merging using the "Squash and merge" button. But basically do whatever works for you best!

lib/req/request.ex
wojtekmach: so to be super clear, I think we'd have eg `Req.get(url, follow_redirects: [location_trusted: true])`. Now, I think what you had initially, `Req.get(url, location_trusted: true)` is actually quite nicer. But let's be consistent with the retry step in the meantime and if we want to change the approach to be more "flat", I'll do it in one go. (Sorry if this seems like a bit of an extra unnecessary work.)

lib/req/steps.ex
srowley: To do this we have to replace the check of the first character in the location string in `put_redirect_location`. I would propose that instead of checking `String.starts_with?(location, "/")` we check `location_url.host` (where `location_url` is the parsed `location`), since the `:host` key of the `location_url` struct would be nil if the location is a relative path.

Alternatively we could pass both the string and the struct to `put_redirect_location` but that seems more messy.

lib/req/request.ex
srowley: This was a little tricky since retry is optional and following redirects is not, but I used the same approach while not providing an option to not follow redirects (that could obviously be added later if needed).

lib/req/steps.ex
srowley: I am just curious as to why one is preferred over the other? The only difference I'm aware of (I think) is that `struct/2` discards keys that aren't in the struct (which doesn't seem to matter one way or another here). Thanks for explaining!

wojtekmach: yeah strucr!/2 would totally work too. Don't have a specific reason, I guess it is personal preference.

lib/req/steps.ex
wojtekmach: > we check `location_url.host`

sounds great!

lib/req/request.ex
wojtekmach: yeah, once we have max_redirects, one way to disable redirects is to set it to 0. `follow_redirects: false` would be another idea. Not sure which one is better. out of the scope of this PR anyway!

wojtekmach:
Thank you!

Issue #14, opened 2021-06-06
`follow_redirects`: Send a GET request after a 301..303


wojtekmach:
Currently we use the method of the original request. See `curl --location`. We should also support all the other 3xx codes provided they have the location header.

thbar:
Just sharing my findings as I wondered about supported redirect codes (308 or not etc):

> We should also support all the other 3xx codes provided they have the location header.

Related:
- https://github.com/benoitc/hackney/issues/635
- https://github.com/edgurgel/httpoison/issues/447

For other readers (such as @fchabouis), req is at time of writing supporting this:

https://github.com/wojtekmach/req/blob/c345e5b5e4b434d7a181ccffd50e5915aecd30f7/lib/req.ex#L829

srowley:
I would be interested in working on this one. Based on what I have found:

- [curl docs](https://curl.se/docs/manpage.html#-L)
- [MDN explanation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections)

it sounds like the desired behavior would be:

- 301..303: redirect, change method to GET
- 307..308: redirect, don't change method

Is that right?

wojtekmach:
@srowley sounds good, please go ahead and thank you!

PR #68, opened 2022-01-18
change method to get on 301..303 redirects

srowley:
Addresses #14

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -624,6 +624,13 @@ defmodule Req.Steps do
   @doc """
   Follows redirects.
 
+  The original request method may be changed to GET depending on the status code:
+
+  | Code          | Method handling    |
+  | ------------- | ------------------ |
+  | 301, 302, 303 | Changed to GET     |
+  | 307, 308      | Method not changed |
+
   ## Options
 
     * `:location_trusted` - by default, authorization credentials are only sent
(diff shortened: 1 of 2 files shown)


lib/req/steps.ex
```diff
@@ -849,6 +857,13 @@ defmodule Req.Steps do
     end
   end
 
+  defp put_redirect_request_method(%Req.Request{method: :get} = request), do: request
```
srowley, line 860: This isn't technically necessary but wasn't sure if it would save an unnecessary update to the struct when the method is already GET.

wojtekmach: I think either way is fine. I'm gonna remove the clause so we have less code around.

lib/req/steps.ex
```diff
@@ -849,6 +857,13 @@ defmodule Req.Steps do
...
+  defp put_redirect_request_method(request) when request.status in 307..308, do: request
+
+  defp put_redirect_request_method(request) do
+    %{request | method: :get}
+  end
```
wojtekmach, line 865:
```suggestion
  defp put_redirect_request_method(request) when request.status in 307..308, do: request

  defp put_redirect_request_method(request), do: %{request | method: :get}
```

wojtekmach:
> Addresses #14

Pro-tip: in the future please write

> Closes #14

This way when the PR is merged, the issue will be automatically closes. Here's a full list of words GitHub detects: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword.

wojtekmach:
Thank you for all the help!

srowley:
Thanks for the tip! My goal with helping is to learn by doing while hopefully contributing in a way that is not more work for you to review than it would be to do these things yourself. If I am not accomplishing that goal at some point I really hope you will let me know.

wojtekmach:
It's definitely no hassle reviewing your PRs. It's a big help and I really appreciate it!

Issue #52, opened 2021-11-10
`auth`: Add `:netrc | {:netrc, path}` as possible values


wojtekmach:
Instead of having a distinct load_netrc step, make it an option to auth, e.g.: `auth: :netrc` which also would be the default?

wojtekmach:
Let's still keep it as a distinct step, it's just auth/2 is going to use it under the hood.

srowley:
Looking at this, I am not sure why keeping `load_netrc` as a step is desirable. Could you confirm?

If you run both steps, it will send two authorization headers, which I believe is not consistent with [RFC7230](https://www.rfc-editor.org/rfc/rfc7230#section-3.2.2).

It makes sense to me to keep `load_netrc` as a private function to run under the hood - maybe this was the intent?

srowley:
Separate question - `load_netrc` seems to be looking for a file with lines like:
```
machine localhost
username foo
password bar
```

But the documentation I can find [here](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html) and [here ](https://everything.curl.dev/usingcurl/netrc)suggests it should be:

```
machine localhost
login foo
password bar
```

or 

`machine localhost login foo password bar`

Should I stick with `username` or use `login`?

wojtekmach:
oops, it should have been `login`, thanks!

PR #71, opened 2022-01-20
Support NETRC environment variable

### LINKED ISSUE #60: `load_netrc`: Support NETRC environment variable
https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html

srowley:
Closes #60.

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -183,6 +183,10 @@ defmodule Req.Steps do
       iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", netrc: "/path/to/custom_netrc").status
       200
 
+      iex> System.put_env("NETRC", "/path/in/environment/variable")
+      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", netrc: :env).status
+      200
+
   """
   @doc step: :request
   def load_netrc(request, path)
@@ -195,6 +199,14 @@ defmodule Req.Steps do
       :error ->
         request
     end
+  end
+
+  def load_netrc(request, :env) do
+    path =
+      System.get_env("NETRC") ||
+        raise "could not find .netrc path, please set the NETRC environment variable"
+
+    load_netrc(request, Path.join(path, ".netrc"))
   end
 
   def load_netrc(request, true) do
--- a/test/req/steps_test.exs
+++ b/test/req/steps_test.exs
@@ -110,13 +110,25 @@ defmodule Req.StepsTest do
 
     assert Req.get!(c.url <> "/auth", netrc: "#{c.tmp_dir}/wrong_netrc").status == 401
 
-    File.write!("#{c.tmp_dir}/correct_netrc", """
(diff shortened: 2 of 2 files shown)


lib/req/steps.ex
```diff
@@ -197,6 +201,14 @@ defmodule Req.Steps do
     end
   end
 
+  def load_netrc(request, :env) do
+    path =
+      System.get_env("NETRC") ||
+        raise "could not find .netrc path, please set the NETRC environment variable"
```
srowley, line 207: This behavior is meant to be consistent with what happens if `System.user_home!/1` is not set.

lib/req/steps.ex
```diff
@@ -183,6 +183,10 @@ defmodule Req.Steps do
       iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", netrc: "/path/to/custom_netrc").status
       200
 
+      iex> System.put_env("NETRC", "/path/in/environment/variable")
+      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", netrc: :env).status
```
wojtekmach, line 187: I think `netrc: :env` makes sense but I feel like when the user already set the environment variable we should take the hint. In other words, we add the `load_netrc` step if `netrc: true | path` or `NETRC` variable is set.

What do you think?

srowley: I am at a disadvantage since I wasn't familiar with `.netrc` until I worked on this, so I don't know, for example, if people might have `NETRC` set but not want to use it for some reason. If it is ok to assume that if it is set is should be used, sure.

If you had `NETRC` set but didn't want to use it, and didn't want to use `.netrc` either, you would have to unset it before making this call. I don't know if that is something someone would ever want to do, or not.

wojtekmach: > I am at a disadvantage since I wasn't familiar with .netrc until I worked on this, so I don't know, for example, if people might have NETRC set but not want to use it for some reason

Right, I think if someone has the file and/or the env variable but they do not want them to be used it is because they want to use _different_ auth credentials. If we today set `netrc: true` and `auth: {user, pass}`, the latter will take precedence because the `auth` step runs _later_.

I had an idea in https://github.com/wojtekmach/req/issues/52#issue-1050348346 that instead of `netrc: true` we should have `auth: :netrc` and as a matter of fact make it the default. I think we should move forward with that idea then.

I think when we have `auth: :netrc`, we try to load `~/.netrc` and if it's not there, we ignore it. We can overwrite the default lookup path with `NETRC` but if we do so and the file is not there, we raise.

WDYT?

Sorry for back'n'forth!

srowley: I would say we just make both changes at the same time as opposed t relying on the default ordering of the steps to always be preserved.  I could just close this PR and combine those two items in a new one; it's pretty minimal work.

wojtekmach: Yeah doing both issues in a single PR sounds good. Feel free to do it here or on a separate one, totally up to you. Thank you!

PR #72, opened 2022-01-22
add :adapter option

### LINKED ISSUE #66: Add `:adapter` option
So that we can conveniently do: `Req.get!(url, adapter: adapter)`. See https://github.com/wojtekmach/req/issues/13#issuecomment-1013979917. Not sure whether this should be on `Req.request` or `Req.Steps.put_default_ateps` though.

srowley:
Closes #66.

Note that the option is added to `Req.build/3`, for a few reasons:

- the adapter runs like a request step, but it is really a required attribute of the request, unlike any other steps
- because the adapter is required and required to run last, it seemed confusing to make changes to something that is required as part of the process of adding defaults. It feels more natural to modify the request defaults when building the request.

I did not add new tests. I did update the two tests that use a custom adapter to add it using this option. Let me know if a separate test is desirable.

### DIFF
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -132,7 +132,7 @@ defmodule Req.Request do
   Sets the request adapter.
 
   Adapter is a request step that is making the actual HTTP request. See
-  `Req.Request.build/3` for more information.
+  `build/3` for more information.
 
   """
   def put_adapter(request, adapter) do
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -118,7 +118,7 @@ defmodule Req.Request do
     :url,
     headers: [],
     body: "",
-    adapter: {Req.Steps, :run_finch, []},
+    adapter: nil,
     unix_socket: nil,
     halted: false,
     request_steps: [],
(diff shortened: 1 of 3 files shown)


lib/req/request.ex
```diff
@@ -118,7 +118,7 @@ defmodule Req.Request do
     :url,
     headers: [],
     body: "",
-    adapter: {Req.Steps, :run_finch, []},
+    adapter: nil,
```
wojtekmach, line 121: wdyt about keeping the default that uses Finch here? If a bit of duplication is the main reason not to do it, I am not too concerned about that at the moment.

srowley: I did this thinking that it is easier to know where to look for the defaults if they are set in one place, but now I see that the struct defaults match the defaults set in `build/3` (I missed this before), so I would say better to keep the default the way it was.

wojtekmach:
Thanks for the PR. If we have the `:adapter` option I wonder if we need the `put_adapter` function anymore, I feel like we don't. WDYT? Let me think about it some more.

srowley:
I don't disagree. If the request needed to be changed after the initial build there is always the option to modify that attribute of the struct directly - it's not like `put_adapter` was performing any validation or other valuable work prior to updating the struct.

wojtekmach:
Thank you!

PR #75, opened 2022-02-06
Replace load_netrc with auth: :netrc | {:netrc, path}

### LINKED ISSUE #52: `auth`: Add `:netrc | {:netrc, path}` as possible values

### LINKED ISSUE #60: `load_netrc`: Support NETRC environment variable

srowley:
Closes #52 and closes #60. This version gets rid of the `load_rtc` step. It uses similar logic under the hood, with the following changes:

1. Look for `login` instead of `username`
2. Parse records stored in `.netrc` as a single line or as multiple lines

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -20,9 +20,6 @@ defmodule Req.Steps do
 
     * [`&put_base_url(&1, options[:base_url])`](`base_url/2`) (if `options[:base_url]` is set)
 
-    * [`&load_netrc(&1, options[:netrc])`](`load_netrc/2`) (if `options[:netrc]` is set
-      to an atom true for default path or a string for custom path)
-
     * [`&auth(&1, options[:auth])`](`auth/2`) (if `options[:auth]` is set)
 
     * [`&put_params(&1, options[:params])`](`put_params/2`) (if `options[:params]` is set)
@@ -51,8 +48,6 @@ defmodule Req.Steps do
 
     * `:base_url` - if set, adds the `put_base_url/2` step
 
-    * `:netrc` - if set, adds the `load_netrc/2` step
-
     * `:auth` - if set, adds the `auth/2` step
 
     * `:params` - if set, adds the `put_params/2` step
@@ -80,7 +75,6 @@ defmodule Req.Steps do
         {Req.Steps, :encode_body, []}
       ] ++
         maybe_steps(options[:base_url], [{Req.Steps, :put_base_url, [options[:base_url]]}]) ++
-        maybe_steps(options[:netrc], [{Req.Steps, :load_netrc, [options[:netrc]]}]) ++
         maybe_steps(options[:auth], [{Req.Steps, :auth, [options[:auth]]}]) ++
         maybe_steps(options[:params], [{Req.Steps, :put_params, [options[:params]]}]) ++
         maybe_steps(options[:range], [{Req.Steps, :put_range, [options[:range]]}]) ++
@@ -145,6 +139,9 @@ defmodule Req.Steps do
 
     * `{username, password}` - uses Basic HTTP authentication
     * `{:bearer, token}` - uses Bearer HTTP authentication
+    * `:netrc` - load credentials from `.netrc` at path specified in `NETRC` environment variable;
+      if `NETRC` is not set, load `.netrc` in user's home directory
+    * `{:netrc, path}` - load credentials from `path`
 
   ## Examples
 
@@ -158,6 +155,12 @@ defmodule Req.Steps do
       iex> Req.get!("https://httpbin.org/bearer", auth: {:bearer, "foo"}).status
       200
 
+      iex> System.put_env("NETRC", "/path/in/environment/variable")
+      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", auth: :netrc).status
+      200
+
+      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", auth: {:netrc, "custom/.netrc"}).status
+      200
   """
   @doc step: :request
   def auth(request, auth)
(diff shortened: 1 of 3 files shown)


lib/req/steps.ex
```diff
@@ -158,6 +155,12 @@ defmodule Req.Steps do
       iex> Req.get!("https://httpbin.org/bearer", auth: {:bearer, "foo"}).status
       200
 
+      iex> System.put_env("NETRC", "/path/in/environment/variable")
```
wojtekmach, line 158:
```suggestion
      iex> System.put_env("NETRC", "./my_netrc")
```

lib/req/steps.ex
```diff
@@ -158,6 +155,12 @@ defmodule Req.Steps do
...
+      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", auth: :netrc).status
+      200
+
+      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", auth: {:netrc, "custom/.netrc"}).status
```
wojtekmach, line 162:
```suggestion
      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", auth: {:netrc, "./my_netrc"}).status
```

lib/req/steps.ex
```diff
@@ -171,23 +174,17 @@ defmodule Req.Steps do
...
-
-      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar").status
-      401
-      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", netrc: true).status
-      200
-      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", netrc: "/path/to/custom_netrc").status
-      200
+  def auth(request, :netrc) do
+    directory = System.get_env("NETRC", System.user_home!())
+    path = Path.join(directory, ".netrc")
```
wojtekmach, line 179: https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html is a vague whether `NETRC` holds the path directory (with `.netrc`) or the path to the file.

I think treating it as the path gives us more flexibility. Seems that's what Requests does too. https://github.com/psf/requests/pull/5643/files#diff-f131fe49192d62f76956ec2fdce8928baaa3cd8cbc070e4d5ac1981642bdadc9R217

```suggestion
    directory = System.get_env("NETRC") || Path.join(System.user_home!(), ".netrc")
```

lib/req/steps.ex
```diff
@@ -197,17 +194,33 @@ defmodule Req.Steps do
     end
   end
 
-  def load_netrc(request, true) do
-    load_netrc(request, Path.join(System.user_home!(), ".netrc"))
+  defp load_netrc(path) do
+    case File.read(path) do
+      {:ok, ""} ->
+        raise ".netrc file is empty."
```
wojtekmach, line 200:
```suggestion
        raise ".netrc file is empty"
```

btw, if we don't have the `{:ok, ""}` clause wouldn't we raise with the `cannot parse .netrc file` exception? If so, let's remove this clause and rely on the latter error instead.

lib/req/steps.ex
```diff
@@ -197,17 +194,33 @@ defmodule Req.Steps do
...
+
+      {:ok, contents} ->
+        contents
+        |> String.trim()
+        |> String.split()
+        |> parse_netrc()
+
+      {:error, reason} ->
+        raise "Error reading .netrc file: #{:file.format_error(reason)}"
```
wojtekmach, line 209:
```suggestion
        raise "error reading .netrc file: #{:file.format_error(reason)}"
```

lib/req/steps.ex
```diff
@@ -197,17 +194,33 @@ defmodule Req.Steps do
...
+  defp parse_netrc(credentials), do: parse_netrc(credentials, %{})
+
+  defp parse_netrc([], acc), do: acc
+
+  defp parse_netrc([_, machine, _, login, _, password | tail], acc) do
+    acc = Map.put(acc, String.trim(machine), {String.trim(login), String.trim(password)})
+    parse_netrc(tail, acc)
   end
 
+  defp parse_netrc(_, _), do: raise("Error parsing .netrc.")
```
wojtekmach, line 222:
```suggestion
  defp parse_netrc(_, _), do: raise("error parsing .netrc file")
```

test/req/steps_test.exs
```diff
@@ -97,36 +111,79 @@ defmodule Req.StepsTest do
       end
     end)
 
-    assert Req.get!(c.url <> "/auth").status == 401
+    directory = System.get_env("NETRC", System.user_home!())
 
-    File.write!("#{c.tmp_dir}/empty_netrc", "")
-    assert Req.get!(c.url <> "/auth", netrc: "#{c.tmp_dir}/empty_netrc").status == 401
+    if File.exists?(Path.join(directory, ".netrc")) do
```
wojtekmach, line 116: this is a bit too clever to have in a test, let's make the test as dumb and deterministic as possible.

test/req/steps_test.exs
```diff
@@ -97,36 +111,79 @@ defmodule Req.StepsTest do
...
     """)
 
-    assert_raise RuntimeError, "parse error: \"bad\"", fn ->
-      Req.get!(c.url <> "/auth", netrc: "#{c.tmp_dir}/bad_netrc")
+    assert_raise RuntimeError, "Error parsing .netrc.", fn ->
+      Req.get!(c.url <> "/auth", auth: {:netrc, "#{c.tmp_dir}/bad_netrc"})
     end
   end
 
+  test "encode_headers/1", c do
```
wojtekmach, line 155: thanks for the extra coverage!

srowley: Sadly that was already there; my changes were so diff-unfriendly that it looks new. :)

lib/req/steps.ex
srowley: Without this clause `parse_netrc` just returns an empty map; it doesn't raise. We could add a clause that catches it and raises; I don't have a preference.

test/req/steps_test.exs
srowley: I agree!

I struggled with this one because without it, the test outcome depends on whether `~/.netrc` is present on the machine running the test. 

Looking at this again I think I can test what I wanted to test with a custom path and that's fine.

wojtekmach:
@srowley you marked a bunch of conversations as resolved but the code remains unchanged. did you forget to push or something? :)

srowley:
I was just using that to keep track of changes I made locally. Sorry if my GitHub etiquette was bad there.

Let me know what you think about the remaining item and then I will push everything.

lib/req/steps.ex
wojtekmach: Lets leave this as is then

wojtekmach:
Stop with the etiquette thing, it is fine. :D

wojtekmach:
this is great, thank you!

srowley:
My pleasure - great learning opportunity for me!

PR #83, opened 2022-03-08
add (first) httpbin.org tests

maxxst:
I thought it might be beneficial to be able to test the library against https://httpbin.org. As an added benefit this enables us to use doctests.

For now I implemented the tests for the
 

- [x] high-level APIs
- [x] auth steps
- [ ] request steps
  - [x] `put_base_url/2`
    - [x] `encode_headers/1`
    - [x] `encode_body/1`
    - [x] `put_params/2`
    - [x] `put_range/2`
    - [x] `put_if_modified_since/2`
    - [ ] `run_steps/2`
- [ ] response steps
  - [x]  `decompress/1`
  - [ ]  `decode_body/1`
- [ ] error steps 

To not interfere with the existing tests and not depend on an internet connection (and a third-party service) I put those tests in an extra `test_httpbin` directory. To run those test just use

```
mix test test_httpbin
```

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -155,11 +155,11 @@ defmodule Req.Steps do
       iex> Req.get!("https://httpbin.org/bearer", auth: {:bearer, "foo"}).status
       200
 
-      iex> System.put_env("NETRC", "./my_netrc")
+      iex> System.put_env("NETRC", "./test/my_netrc")
       iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", auth: :netrc).status
       200
 
-      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", auth: {:netrc, "./my_netrc"}).status
+      iex> Req.get!("https://httpbin.org/basic-auth/foo/bar", auth: {:netrc, "./test/my_netrc"}).status
       200
   """
   @doc step: :request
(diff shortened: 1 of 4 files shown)


test_httpbin/steps/request_steps_test.exs
```diff
@@ -0,0 +1,13 @@
...
+  doctest Req.Steps,
+    only: [
+      put_base_url: 2,
+      encode_headers: 1,
+      encode_body: 1,
+      put_params: 2,
+      put_range: 2,
+      put_if_modified_since: 2
+    ]
+end
```
wojtekmach, line 13: Could you do the following changes?

1. Move all `doctest` declarations to a single case: `test/httpbin_test.exs`
2. Add `@moduletag :httpbin` to it
3. in test/helper run `unless System.get_env("CI") do ExUnit.configure(exclude: :httpbin) end`

WDYT?

wojtekmach (commented):
This is great. Always felt a bit bad about not running these. Left some comments below!

test_httpbin/req_test.exs
```diff
@@ -0,0 +1,79 @@
+defmodule ReqTest do
```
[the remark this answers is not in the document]
maxxst: What do you think about keeping it but skipping by default ? Aren't more tests better? Just to minimize the chance to run into a bypass error ...

test_httpbin/steps/request_steps_test.exs
maxxst: regarding 1. I personally like more separation of different tests but if you prefer a single testfile I can update that this way.

test_httpbin/req_test.exs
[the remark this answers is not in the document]
wojtekmach: No, I believe more tests ain't better. It's more stuff to maintain. Let's remove this for now, we can always revisit it in the future.

test_httpbin/steps/request_steps_test.exs
wojtekmach: Let's start with a single test case and see how it goes, we can always break it apart down the road if it's weird.

maxxst:
I think now it should be according to your plan.
The following command will test everything including the httpbin doctests
```
mix test --include httpbin
```

wojtekmach:
Thank you!

wojtekmach:
FYI here's the patch to run doctests for run_steps. :) f10b0f8

Issue #84, opened 2022-03-28
`put_base_url/2` would raise an exception when `retry`


dsdshcym:
Hi @wojtekmach,
thanks for this great library, we've been using it in production for a while without any issues!

But today I want to use the `retry` step, but found that `put_base_url/2` would raise an exception here, stopping `retry` from sending the request:
https://github.com/wojtekmach/req/blob/f10b0f87a906764d485aeed1272344a5804b6db7/lib/req/steps.ex#L124-L126

Becuase the `url` was set in the first attempt,
the `match?(%{scheme: nil, host: nil}, request.url)` test would not pass,
thus the second and later attempts would always fail here.

I think a proper fix is to remove this check here,
as an user, I'd already expect `put_base_url` to replace the base_url anyway (because it's a `put_*` function)

dsdshcym:
ping @wojtekmach in case you missed this issue 😅

PR #85, opened 2022-03-28
Overwrite base_url in put_base_url/2, instead of raise an error

dsdshcym:
- fixes https://github.com/wojtekmach/req/issues/84

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -109,6 +109,8 @@ defmodule Req.Steps do
   @doc """
   Sets base URL for all requests.
 
+  Note this step would overwrite existing base_url, see examples below:
+
   ## Examples
 
       iex> options = [base_url: "https://httpbin.org"]
@@ -116,14 +118,12 @@ defmodule Req.Steps do
       200
       iex> Req.get!("/status/201", options).status
       201
+      iex> Req.get!("https://unknown-host/status/201", options).status
+      201
   """
   @doc step: :request
   def put_base_url(request, base_url) when is_binary(base_url) do
     # TODO: change build/3 so that the url is parsed later so that here it is not yet parsed
-
-    unless match?(%{scheme: nil, host: nil}, request.url) do
-      raise "put_base_url/2 expects the request url to only contain a path, got: #{URI.to_string(request.url)}"
-    end
 
     # remove when we require Elixir v1.13
     url = request.url.path || ""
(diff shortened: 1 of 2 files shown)


lib/req/steps.ex
```diff
@@ -109,22 +109,22 @@ defmodule Req.Steps do
...
       201
+      iex> Req.get!("https://unknown-host/status/201", options).status
+      201
   """
   @doc step: :request
   def put_base_url(request, base_url) when is_binary(base_url) do
     # TODO: change build/3 so that the url is parsed later so that here it is not yet parsed
 
-    unless match?(%{scheme: nil, host: nil}, request.url) do
-      raise "put_base_url/2 expects the request url to only contain a path, got: #{URI.to_string(request.url)}"
```
dsdshcym, line 125: maybe log a warning here?

lib/req/steps.ex
```diff
@@ -109,22 +109,22 @@ defmodule Req.Steps do
...
+  Note this step would overwrite existing base_url, see examples below:
+
   ## Examples
 
       iex> options = [base_url: "https://httpbin.org"]
       iex> Req.get!("/status/200", options).status
       200
       iex> Req.get!("/status/201", options).status
       201
+      iex> Req.get!("https://unknown-host/status/201", options).status
```
wojtekmach, line 121: I don't think we should ignore `https://unknown-host` that the user just given. If anything, I think it should take precedence over `:base_url`. That's what Tesla [1] does too. If we have that behaviour, would that be OK for your use case?

[1] https://github.com/teamon/tesla/blob/v1.4.4/lib/tesla/middleware/core.ex#L35

dsdshcym: Yes, I agree we shouldn't ignore the URL given by the user

But I'm not sure if silently using that over `:base_url` is a good idea
I'd prefer to raise in this case (which is exactly the old behavior I'm changing here 😂 )

Maybe we should distinguish the `base_url + path` case and the `full_url` case in the `Req` struct?
So that we can raise when `full_url` is set and `put_base_url` gets called, and only override `base_url` when `full_url` is nil

If this sounds too complicated to you, I'm okay with taking `full_url` over `base_url`, as in that case, `retry` would still work with `put_base_url` (#84)

wojtekmach: > I'm okay with taking full_url over base_url

Yup, let's roll with that!

PR #88, opened 2022-04-22
New API

wojtekmach:
Major changes:

  * Add high-level functional API: `Req.new(options) |> Req.request(options)`

  * Add `Req.Request.options` field that steps can read from. Also, make
    all steps be arity 1.

    When using "High-level" API, we now run all steps by default. (The
    steps, by looking at `request.options`, can decide to be no-op.)

  * Remove `put_default_steps/2` step.

  * Remove `run_steps/2` step.

Deprecations:

  * Deprecate calling `Req.post!(url, body)` in favour of `Req.post(url,
    body: body)`. Same for `Req.put!`.

  * Deprecate setting `retry: [delay: delay, max_retries: max_retries]`
    in favour of `[retry_delay: delay, max_retries: max_retries]`.

  * Deprecate setting `follow_redirects: [location_trusted: trusted]`
    in favour of `[location_trusted: trusted]`.

  * Deprecate setting `cache: [dir: dir]` in favour of `cache_dir: dir`.

  * Deprecate `Req.Request.build/3` in favour of `Req.Request.new/1`.

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -334,15 +334,15 @@ defmodule Req do
   def request(request, options) do
     options = Keyword.merge(default_options(), options)
 
-      {request_options, options} = Keyword.split(options, [:method, :url, :headers, :body])
+    {request_options, options} = Keyword.split(options, [:method, :url, :headers, :body])
 
-      request =
-        Map.merge(request, Map.new(request_options), fn
-            :url, _, url -> URI.parse(url)
-            # TODO: merge headers
-            :headers, _, headers -> headers
-            _, _, value -> value
-        end)
+    request =
+      Map.merge(request, Map.new(request_options), fn
+        :url, _, url -> URI.parse(url)
+        # TODO: merge headers
+        :headers, _, headers -> headers
+        _, _, value -> value
+      end)
 
     request = update_in(request.options, &Map.merge(&1, Map.new(options)))
     Req.Request.run(request)
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -10,46 +10,176 @@ defmodule Req do
 
   @type method() :: :get | :post | :put | :delete
 
+  @default_options %{
+    # request steps
+    auth: nil,
+    base_url: nil,
+    cache: nil,
+    cache_dir: nil,
+    params: [],
+    range: nil,
+    finch: Req.Finch,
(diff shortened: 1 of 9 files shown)


lib/req.ex
```diff
@@ -10,92 +10,314 @@ defmodule Req do
...
+              end
+
+            # TODO: merge headers
+
+            {%{request | option => value}, Keyword.delete(options, option)}
+
+          :error ->
+            {request, options}
+        end
+      end)
```
jonatanklosko, line 281:
```suggestion
      {request_options, options} = Keyword.split(options, [:method, :url, :headers, :body])

      request =
        Map.merge(request, Map.new(request_options), fn
            :url, _, url -> URI.parse(url)
            # TODO: merge headers
            :headers, _, headers -> headers
            _, _, value -> value
        end)
```

lib/req.ex
```diff
@@ -10,92 +10,314 @@ defmodule Req do
...
+    # response steps
+    follow_redirects: true,
+    location_trusted: false,
+    raw: false,
+
+    # error steps
+    max_retries: 2,
+    retry: true,
+    retry_delay: 2000
+  }
```
jonatanklosko, line 34: Since we include all the steps, why can't this be empty and make each step get its options with the corresponding defaults?

jonatanklosko:
Exciting! :cat:

lib/req.ex
wojtekmach: Good question! I think that's actually preferable but by supplying defaults we can use guards: https://github.com/wojtekmach/req/pull/88/files#diff-07dc6fd4ec95415d610a3e3ddf9374540d6d64d02eaf42df7a233d6f05b32cc1R425. But it's probably not worth it just for guards sake, dunno. Will think about it some more.

wojtekmach: oops, link doesn't work, here's an example:

```elixir
  def decompress({request, response}) when request.options.raw == true do
     {request, response}
   end
```

jonatanklosko: If we dropped the `raw: false` default the same guard would work though, no?

wojtekmach: oh, guards shouldn't fail if `options` doesn't have `:raw` key. Yup, thanks!

PR #92, opened 2022-04-25
Add `:output` step

### LINKED ISSUE #21: New step: `output`
```
% curl -o foo.html https://elixir-lang.org/index.html
% curl --remote-name https://elixir-lang.org/index.html
```

Let's add:

```elixir
iex> Req.get!(url, output: path)
iex> Req.get!(url, output: :remote_name)
```

srowley:
Closes #21.

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -84,6 +84,7 @@ defmodule Req do
         &Req.Steps.retry/1,
         &Req.Steps.follow_redirects/1,
         &Req.Steps.decompress/1,
+        &Req.Steps.output/1,
         &Req.Steps.decode_body/1
       ],
       error_steps: [
@@ -275,8 +276,13 @@ defmodule Req do
   Response body options:
 
     * `:raw` - if set to `true`, disables automatic body decompression
-      ([`decompress`](`Req.Steps.decompress/1`) step) and and decoding
+      ([`decompress`](`Req.Steps.decompress/1`) step) and decoding
       ([`decode_body`](`Req.Steps.decode_body/1`) step). Defaults to `false`.
+
+    * `:output` - if set to `:remote_name`, saves the response body (after decompression
+      but before decoding) to a file with the same name as the remote file. If set to a
+      path name, saves the the response body at the given path. Once the file is written,
+      the response body is replaced with an empty string.
 
   Response redirect options ([`follow_redirects`](`Req.Steps.follow_redirects/1`) step):
 
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -279,10 +279,10 @@ defmodule Req do
       ([`decompress`](`Req.Steps.decompress/1`) step) and decoding
       ([`decode_body`](`Req.Steps.decode_body/1`) step). Defaults to `false`.
 
-    * `:output` - if set to `:remote_name`, saves the response body (after decompression
-      but before decoding) to a file with the same name as the remote file. If set to a
-      path name, saves the the response body at the given path. Once the file is written,
-      the response body is replaced with an empty string.
+    * `:output` - if set, writes the response body to a file (via
+      [`output`](`Req.Steps.output/1`)). Can be set to a string path or an atom `:remote_name` which
+      would use the remote name as the filename in the current working directory. Once the file is
+      written, the response body is replaced with `""`.
 
   Response redirect options ([`follow_redirects`](`Req.Steps.follow_redirects/1`) step):
 
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -611,6 +611,40 @@ defmodule Req.Steps do
   end
 
   @doc """
+  Writes the response body to a file.
+
+  This step runs after the ([`decompress`](`Req.Steps.decompress/1`) step) and before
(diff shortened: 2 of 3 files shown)


lib/req.ex
```diff
@@ -274,9 +275,14 @@ defmodule Req do
...
 
     * `:raw` - if set to `true`, disables automatic body decompression
-      ([`decompress`](`Req.Steps.decompress/1`) step) and and decoding
+      ([`decompress`](`Req.Steps.decompress/1`) step) and decoding
       ([`decode_body`](`Req.Steps.decode_body/1`) step). Defaults to `false`.
 
+    * `:output` - if set to `:remote_name`, saves the response body (after decompression
+      but before decoding) to a file with the same name as the remote file. If set to a
+      path name, saves the the response body at the given path. Once the file is written,
+      the response body is replaced with an empty string.
```
wojtekmach, line 284:
```suggestion
    * `:output` - if set, writes the response body to a file (via
      [`output`](`Req.Steps.output/1`)). Can be set to a String path or an atom `:remote_name` which
      would use the remote name as the filename in the current working directory. Once the file is
      written, the response body is replaced with `""`.
```

lib/req/steps.ex
```diff
@@ -534,6 +534,40 @@ defmodule Req.Steps do
...
+  the ([`decode_body`](`Req.Steps.decode_body/1`) step). After the output file is written,
+  the response body is set to `""`.
+
+  ## Examples
+
+      # Saves file to specified path
+      Req.get!("https://elixir-lang.org/index.html", output: "elixir_home.html")
+
+      # Saves file to current working directory with same name as remote file
+      Req.get!("https://elixir-lang.org/index.html", output: :remote_name)
```
wojtekmach, line 550:
```suggestion
  After the output file is written, the response body is set to `""`.

  ## Request Options

    * `:output` - if set, writes the response body to a file. Can be one of:

      * `path` - writes to the given path

      * `:remote_name` - uses the remote name as the filename in the current working directory

  ## Examples

      iex> Req.get!("https://elixir-lang.org/index.html", output: "/tmp/elixir_home.html")
      iex> File.exists?("/tmp/elixir_home.html")
      true

      iex> Req.get!("https://elixir-lang.org/index.html", output: :remote_name)
      iex> File.exists?("index.html")
      true
```

test/req/steps_test.exs
```diff
@@ -300,6 +300,35 @@ defmodule Req.StepsTest do
...
+
+  @tag :tmp_dir
+  test "output/1: :remote_name (not compressed)", c do
+    Bypass.expect_once(c.bypass, "GET", "/directory/save_this.txt", fn conn ->
+      Plug.Conn.send_resp(conn, 200, "to be filed")
+    end)
+
+    cwd = File.cwd!()
+    on_exit(fn -> File.cd!(cwd) end)
+    File.cd!(c.tmp_dir)
```
wojtekmach, line 324: unfortunately if we do this, we should make the whole test case `async: false` as sooner or later we'll run into a problem.

I'd do this instead:

```elixir
test "..." do
  response = Req.get!(c.url <> "/directory/save_this.txt", output: :remote_name)
  assert response.body == ""
  assert File.read!("save_this.txt") == "to be filed"
after
  File.rm("save_this.txt")
end
```

test/req/steps_test.exs
```diff
@@ -300,6 +300,35 @@ defmodule Req.StepsTest do
...
+
+    response = Req.get!(c.url <> "/directory/save_this.txt", output: :remote_name)
+
+    assert response.body == ""
+    assert File.read!(c.tmp_dir <> "/save_this.txt") == "to be filed"
+  end
```
wojtekmach, line 330: we also need to add a test for empty filename. We should raise like curl does:

```
$ curl --remote-name https://elixir-lang.org
curl: Remote file name has no length!
curl: (23) Failed writing received data to disk/application

$ curl --remote-name https://elixir-lang.org/
curl: Remote file name has no length!
curl: (23) Failed writing received data to disk/application

$ curl -o "" https://elixir-lang.org
Warning: Remote filename has no length!
curl: (23) Failure writing output to destination
```

perhaps our message would be something like: `cannot write to file ""` or something, up to you.

wojtekmach (commented):
Great, I left a few comments below.

test/req/steps_test.exs
srowley: I thought about that but thought about it wrong; you are right. I didn't realize you could use `after` this way; that will help me elsewhere; thanks!

wojtekmach:
Excellent, thank you!

wojtekmach:
@srowley just fyi, I ended up deprecating (and soon removing) `:output` in favour of [`into: File.stream!(path)`](https://github.com/wojtekmach/req/blob/d161b898c67af74e63fdfbba8fb0aa83abfc0844/CHANGELOG.md#add-response-body-streaming) in #231. We don't have an equivalent of `output: :remote_name` but that's ok for now, can always add it later. Thanks for working on this and other features!

srowley:
That's very kind and appreciate the thought - it's my pleasure/privilege to contribute!

PR #93, opened 2022-04-27
Add Req.Request.prepare/1

### LINKED ISSUE #69: Add `Request.prepare`
This function will return a request that executed all the request steps. The response and error steps are yet to be executed of course.

Example:

```elixir
iex> r = Req.prepare(method: :get, url: "http://example.com", params: [x: 1], auth: {"foo", "bar"})
iex> r.url
"http://example.com/?x=1"
iex> r.headers
[{"authorization", "Basic " <> ...}]
iex> Req.Request.run!(r)
%Req.Response{...}
```

srowley:
Closes #69.

I wasn't sure if `prepare/1` should be required to handle a halted request. It doesn't appear to me that a pipeline can be halted before all of the request steps have run, so I didn't handle it for now.

### DIFF
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -261,6 +261,43 @@ defmodule Req.Request do
   end
 
   @doc """
+  Runs the request steps in the request pipeline but does not execute them.
+
+  Accepts a `Req.Request` struct, or the same keyword list of options accepted
+  by `Req.new/1`.
+
+  Returns a `Req.Request` struct with all of its steps executed.
+
+  ## Examples
+
+      iex> prepared = Req.Request.prepare(method: :get, base_url: "http://www.example.com", params: [x: 1])
+      iex> prepared.url.host
+      "www.example.com"
+      iex> prepared.url.query
+      "x=1"
+  """
+  def prepare(options) when is_list(options) do
+    options
+    |> Req.new()
+    |> prepare()
+  end
+
+  def prepare(%Req.Request{request_steps: []} = request) do
+    request
+  end
+
+  def prepare(%{request_steps: [step | steps]} = request) do
+    case run_step(step, request) do
+      %Req.Request{} = request ->
+        request = %{request | request_steps: steps}
+        prepare(request)
+
+      {_request, %{__exception__: true} = exception} ->
+        raise exception
+    end
+  end
+
+  @doc """
   Runs a request pipeline.
 
   Returns `{:ok, response}` or `{:error, exception}`.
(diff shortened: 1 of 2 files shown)


lib/req/request.ex
```diff
@@ -260,6 +260,43 @@ defmodule Req.Request do
...
+    request
+  end
+
+  def prepare(%{request_steps: [step | steps]} = request) do
+    case run_step(step, request) do
+      %Req.Request{} = request ->
+        request = %{request | request_steps: steps}
+        prepare(request)
+
+      {_request, %{__exception__: true} = exception} ->
```
wojtekmach, line 295: please add a test for this clause. Also, I'm not sure if we should raise the exception we got, I feel like we should raise an exception saying that we cannot prepare a request with a step that returns an exception and then print the exception. Similarly, if a request step returns {request, response}, we should raise too.

lib/req/request.ex
```diff
@@ -260,6 +260,43 @@ defmodule Req.Request do
...
+  Returns a `Req.Request` struct with all of its steps executed.
+
+  ## Examples
+
+      iex> prepared = Req.Request.prepare(method: :get, base_url: "http://www.example.com", params: [x: 1])
+      iex> prepared.url.host
+      "www.example.com"
+      iex> prepared.url.query
+      "x=1"
+  """
```
wojtekmach, line 278: `Req` calls functions from `Req.Request` but it shouldn't be the other way around. So let's not accept options, let's just accept a request struct.

```suggestion
  @doc """
  Prepares the request by running all of its request steps.

  ## Examples

      iex> req = Req.new(base_url: "https://httpbin.org", url: "/anything")
      iex> req = Req.Request.prepare(req)
      iex> to_string(req.url)
      "https://httpbin.org/anything"
  """
```

wojtekmach:
Thank you @srowley! I've made some changes on main but primarily I decided to hide this function for now. It is going to be extremely useful internally, it's gonna simplify a bunch of request step tests as for most of them we don't even need to _make_ the request (and thus have bypass and all) and instead can just inspect the url, the headers, etc. I'm hesitant to expose this function publicly because I don't know how we should handle things like:

```elixir
Req.new() |> Req.Request.prepare() |> Req.get!(auth: {"foo", "bar"})
```

that is we try to use options that are reserved for request steps on a prepared request. I suppose we could store a `request.private.req_prepared = true` flag and then check it in `Req.request/1` but I'm not yet sure about unintended consequences so better to wait up. Again, thanks for the PR!

srowley:
Great! Sorry I had not yet addressed your comments; I was about half done and then didn't have time for awhile. If at some point you would like to those changes to be made let me know.

PR #96, opened 2022-05-04
add compress_request step

### LINKED ISSUE #25: New step: `compress`
This compresses the _request_ body. Gzips the request body and sets the appropriate request header.

Usage:

```elixir
Req.post!(url, body: "foo", compress: :gzip)
```

srowley:
Closes #25.

I took liberties with the name as I thought `compress_request` was clearer.

I could not come up with a good example for the docs without using `prepare/1`. Unfortunately compressed requests to [httpbin.org are not decoded due to security risks](https://github.com/postmanlabs/httpbin/issues/577).

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -77,7 +77,7 @@ defmodule Req do
         put_range: &Req.Steps.put_range/1,
         cache: &Req.Steps.cache/1,
         put_plug: &Req.Steps.put_plug/1,
-        compress_request: &Req.Steps.compress_request/1
+        compress_body: &Req.Steps.compress_body/1
       ],
       response_steps: [
         retry: &Req.Steps.retry/1,
@@ -362,8 +362,8 @@ defmodule Req do
     * `:body` - the request body. The body is automatically encoded using the
       [`encode_body`](`Req.Steps.encode_body/1`) step.
 
-    * `:compress_request` - compresses the request body via the
-      [`compress_request`](`Req.Steps.compress_request/1`) step.
+    * `:compress_body` - compresses the request body via the
+      [`compress_body`](`Req.Steps.body/1`) step.
 
   Additional URL options:
 
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -76,7 +76,8 @@ defmodule Req do
         put_params: &Req.Steps.put_params/1,
         put_range: &Req.Steps.put_range/1,
         cache: &Req.Steps.cache/1,
-        put_plug: &Req.Steps.put_plug/1
+        put_plug: &Req.Steps.put_plug/1,
+        compress_request: &Req.Steps.compress_request/1
       ],
       response_steps: [
         retry: &Req.Steps.retry/1,
@@ -360,6 +361,9 @@ defmodule Req do
 
     * `:body` - the request body. The body is automatically encoded using the
       [`encode_body`](`Req.Steps.encode_body/1`) step.
+
+    * `:compress_request` - compresses the request body via the
+      [`compress_request`](`Req.Steps.compress_request/1`) step.
 
   Additional URL options:
 
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -363,7 +363,7 @@ defmodule Req do
       [`encode_body`](`Req.Steps.encode_body/1`) step.
 
     * `:compress_body` - compresses the request body via the
-      [`compress_body`](`Req.Steps.body/1`) step.
+      [`compress_body`](`Req.Steps.compress_body/1`) step.
 
   Additional URL options:
 
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -377,17 +377,17 @@ defmodule Req.Steps do
   @doc """
   Compresses the request body.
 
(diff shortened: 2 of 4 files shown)


lib/req/steps.ex
```diff
@@ -374,6 +374,26 @@ defmodule Req.Steps do
     end
   end
 
+  @doc """
+  Compresses the request body.
+
+  ## Request options
```
wojtekmach, line 380:
```suggestion
  ## Request Options
```

lib/req/steps.ex
```diff
@@ -374,6 +374,26 @@ defmodule Req.Steps do
...
+
+    * `:compress_request` - compresses the request body using `gzip` compression and sets the
+    `"Content-Encoding: gzip"` header. Defaults to `false`.
+
+  """
+  @doc step: :request
+  def compress_request(request) do
+    if request.options[:compress_request] do
+      request
+      |> Map.put(:body, :zlib.gzip(request.body))
```
wojtekmach, line 390:
```suggestion
      |> Map.update!(:body, &:zlib.gzip/1)
```

lib/req/steps.ex
```diff
@@ -374,6 +374,26 @@ defmodule Req.Steps do
...
+  Compresses the request body.
+
+  ## Request options
+
+    * `:compress_request` - compresses the request body using `gzip` compression and sets the
+    `"Content-Encoding: gzip"` header. Defaults to `false`.
+
+  """
+  @doc step: :request
+  def compress_request(request) do
```
wojtekmach, line 387: Given we have `decompress` response step, let's call this one `compress`.

If you think a more verbose name would be better, let's call this `compress_body` then? (Since we have `decompress_body` response step.)

```suggestion
  def compress(request) do
```

lib/req/steps.ex
```diff
@@ -374,6 +374,26 @@ defmodule Req.Steps do
...
+    if request.options[:compress_request] do
```
wojtekmach, line 388: 1. let's call the option `:compress` (even if we're gonna end up renaming the step)

2. let's support other compression algorithms two (well, only one more for now):

```elixir
case Map.fetch(request.options, :compress) do
  {:ok, :gzip} ->
    request
    |> Map.update!(:body, &:zlib.gzip/1)
    |> put_header

  {:ok, :deflate} ->
    request
    |> Map.update!(:body, &:zlib.zip/1)
    |> put_header

  {:ok, algorithm} ->
    raise "expected :compress to be :gzip or :deflate, got: #{inspect(algorithm)}"

  :error ->
    request
end
```

wojtekmach (commented):
Looks good, a few comments below. Please also mention the step in the README.

lib/req/steps.ex
wojtekmach: Please ignore this, compress: true is fine for now.

lib/req/steps.ex
srowley: Let's do `compress_body` if you agree.. Having `compress` and `compressed` is a big opportunity for confusion. It also makes it hard to search through the code.

lib/req/steps.ex
srowley: Will do but to be honest I think clarifying what is being compressed is helpful here also.

lib/req/steps.ex
wojtekmach: If we don’t have compressed would you be ok with calling it just compress? I agree about possible confusion

srowley: I suppose it depends on what we have instead but I assume it would.

wojtekmach: @srowley sorry for back'n'forth and bike shedding. Let's go with your idea of `compress_body/1` and `:compress_body` option.

I tend to prefer shorter option names but this would be rather seldom used one so it's ok if it's a bit longer. It's definitely super clear. I'll probably keep the name `compressed` though since there's `curl --compressed` and can't think of a better one.

If you'd like to send a PR to rename decompress to decompress_body that would be appreciated too but no worries, happy to do it myself once this lands. Thank you.

lib/req.ex
```diff
@@ -361,6 +362,9 @@ defmodule Req do
     * `:body` - the request body. The body is automatically encoded using the
       [`encode_body`](`Req.Steps.encode_body/1`) step.
 
+    * `:compress_body` - compresses the request body via the
+      [`compress_body`](`Req.Steps.body/1`) step.
```
wojtekmach, line 366:
```suggestion
      [`compress_body`](`Req.Steps.compress_body/1`) step.
```

lib/req/steps.ex
```diff
@@ -374,6 +374,26 @@ defmodule Req.Steps do
     end
   end
 
+  @doc """
+  Compresses the request body.
+
+  ## Request Options
+
+    * `:compress_body` - compresses the request body using `gzip` compression and sets the
```
wojtekmach, line 382:
```suggestion
    * `:compress_body` - compresses the request body using gzip and sets the
```

wojtekmach:
Thank you!

PR #104, opened 2022-05-11
support brotli decompression

### LINKED ISSUE #101: `decompress_body`: Support brotli
Use https://hex.pm/packages/brotli

srowley:
Closes #101.

I assumed that if we were going to decompress with this algorithm we also needed to support requesting in `compress/1`.


lib/req/steps.ex
```diff
@@ -558,6 +577,14 @@ defmodule Req.Steps do
   @doc """
   Decompresses the response body based on the `content-encoding` header.
 
+  Supported formats:
+
+  | Header        | Decompression Algorithm                         |
```
wojtekmach, line 582:
```suggestion
  | Value         | Decompression Algorithm                         |
```

lib/req/steps.ex
```diff
@@ -558,6 +577,14 @@ defmodule Req.Steps do
...
+  | ------------- | ------------------------------------------------|
+  | gzip, x-gzip  | `:zlib.gunzip/1`                                |
+  | zip           | `:zlib.unzip/1`                                 |
+  | br            | `:brotli.decode/1` (if `:brotli` is installed) |
```
wojtekmach, line 586:
```suggestion
  | br            | `:brotli.decode/1` (if `:brotli` is installed)  |
```

wojtekmach:
This is fantastic, thank you! I added minor improvements on main: cd77a46

PR #105, opened 2022-05-11
Add `Req.Request.put_header/3` function

aleDsz:
It implements the same behaviour as [`Plug.Conn.put_req_header/3`](https://hexdocs.pm/plug/Plug.Conn.html#put_req_header/3) does.

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -417,14 +417,12 @@ defmodule Req do
         * atom header names are turned into strings, replacing `-` with `_`. For example,
           `:user_agent` becomes `"user-agent"`
 
-        * string header names are left as is
+        * string header names are left as is. Because header keys are case-insensitive
+          in both HTTP/1.1 and HTTP/2, it is recommended for header keys to be in
+          lowercase, to avoid sending duplicate keys in a request.
 
         * `NaiveDateTime` and `DateTime` header values are encoded as "HTTP date". Otherwise,
           the header value is encoded with `String.Chars.to_string/1`.
-
-        * Because header keys are case-insensitive in both HTTP/1.1 and HTTP/2,
-          it is recommended for header keys to be in lowercase, to avoid sending
-          duplicate keys in a request.
 
       If you set `:headers` options both in `Req.new/1` and `request/2`, the header lists are merged.
 
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -422,6 +422,10 @@ defmodule Req do
         * `NaiveDateTime` and `DateTime` header values are encoded as "HTTP date". Otherwise,
           the header value is encoded with `String.Chars.to_string/1`.
 
+        * Because header keys are case-insensitive in both HTTP/1.1 and HTTP/2,
+          it is recommended for header keys to be in lowercase, to avoid sending
+          duplicate keys in a request.
+
       If you set `:headers` options both in `Req.new/1` and `request/2`, the header lists are merged.
 
     * `:body` - the request body.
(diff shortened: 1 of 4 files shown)


lib/req/steps.ex
```diff
@@ -227,10 +227,10 @@ defmodule Req.Steps do
   def compressed(request) do
     case Map.fetch(request.options, :compressed) do
       :error ->
-        put_new_header(request, "accept-encoding", supported_accept_encoding())
+        Req.Request.put_header(request, "accept-encoding", supported_accept_encoding())
```
wojtekmach, line 230: there is an important difference between put_new_header and the new function. With the latter, it would be impossible for the user to set accept-encoding header themselves, it would always be overridden by the compressed step. I think that's undesirable.

We could add a Req.Request.put_new_header but let's hold off for more use cases. For now, please revert this and similar changes.

wojtekmach: Actually please add put_new_header but add doc: false

lib/req/steps.ex
```diff
@@ -1135,15 +1135,11 @@ defmodule Req.Steps do
   ## Utilities
 
   defp put_new_header(struct, name, value) do
```
wojtekmach, line 1137: all this function does is calling Req.Request.put_new_header. Let's remove this indirection and call the function directly at call sites

lib/req/steps.ex
```diff
@@ -1135,15 +1135,11 @@ defmodule Req.Steps do
...
-    if Enum.any?(struct.headers, fn {key, _} -> String.downcase(key) == name end) do
-      struct
-    else
-      put_header(struct, name, value)
-    end
+    Req.Request.put_new_header(struct, name, value)
   end
 
   defp put_header(struct, name, value) do
```
wojtekmach, line 1141: same here.

lib/req/request.ex
```diff
@@ -470,6 +470,73 @@ defmodule Req.Request do
...
+      ** (RuntimeError) put_header: header key is not lowercase: "Accept"
+
+  """
+  @spec put_header(t(), binary(), binary()) :: t()
+  def put_header(request, key, value)
+
+  def put_header(%Req.Request{halted: true}, _key, _value) do
+    raise "put_header: request is already halted"
+  end
+
```
wojtekmach, line 506: let's remove this for now since we don't do such check anywhere else.

```suggestion
```

lib/req/request.ex
```diff
@@ -470,6 +470,73 @@ defmodule Req.Request do
...
+  def put_header(%Req.Request{headers: headers} = request, key, value)
+      when is_binary(key) and is_binary(value) do
+    ensure_lowercase_header_key!(key)
```
wojtekmach, line 509: I think it is a good idea to validate header names but let's hold off on it. Perhaps we'd someday make a step and people could opt-in to do that. We basically follow Plug which doesn't enforce.

```suggestion
```

lib/req.ex
```diff
@@ -422,6 +422,10 @@ defmodule Req do
         * `NaiveDateTime` and `DateTime` header values are encoded as "HTTP date". Otherwise,
           the header value is encoded with `String.Chars.to_string/1`.
 
+        * Because header keys are case-insensitive in both HTTP/1.1 and HTTP/2,
```
wojtekmach, line 425: please move this advice to `  * string header names are left as is` list item.

wojtekmach:
Thank you!

PR #108, opened 2022-05-18
Relax finch version requirement to allow 0.12

bsedat:
See the associated [Finch Changelog](https://github.com/sneako/finch/blob/f07ef3063b2778776ea847bab45fc7b566d02668/CHANGELOG.md). It appears that the only breaking changes are to do with Telemetry, which I don't believe Req interacts with currently.
Finch 0.12.0 also adds non-GET request support when using an HTTP proxy.


wojtekmach:
Thank you!

PR #110, opened 2022-05-22
Handle 429 Retry-After header

### LINKED ISSUE #43: `retry`: Handle 429 Retry-After header
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

srowley:
Closes #43.

I don't love how this is tested, so if you know of a better approach I am more than happy to implement it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -310,8 +310,10 @@ See "Adapter" seciton in `Req.Request` module documentation for more information
   * `follow_redirects`: Don't send auth headers on redirect to different scheme/host/port
     unless `location_trusted: true` is set
 
+  * `retry`: The `:retry` option handles the `Retry-After` response header on HTTP 429 responses 
+
   * `retry`: The `:retry` option can now be set to `:safe` (default) to only retry GET/HEAD
-    requests on HTTP 408/5xx responses or exceptions, `:always` to always retry, `:never` to never
+    requests on HTTP 408/429/5xx responses or exceptions, `:always` to always retry, `:never` to never
     retry, and `fun` - a 1-arity function that accepts either a `Req.Response` or an exception
     struct and returns boolean whether to retry
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -310,7 +310,7 @@ See "Adapter" seciton in `Req.Request` module documentation for more information
   * `follow_redirects`: Don't send auth headers on redirect to different scheme/host/port
     unless `location_trusted: true` is set
 
-  * `retry`: The `:retry` option handles the `Retry-After` response header on HTTP 429 responses 
+  * `retry`: The `Retry-After` response header on HTTP 429 responses is now respected
 
   * `retry`: The `:retry` option can now be set to `:safe` (default) to only retry GET/HEAD
     requests on HTTP 408/429/5xx responses or exceptions, `:always` to always retry, `:never` to never
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -1050,8 +1050,8 @@ defmodule Req.Steps do
           and returns boolean whether to retry
 
     * `:retry_delay` - sleep this number of milliseconds before making another attempt, defaults
-      to `#{@default_retry_delay}`. Will be ignored if HTTP 429 response includes `Retry-After`
-      header, in favor of the delay interval specified in the header.
+      to `#{@default_retry_delay}`. If the response is HTTP 429 and contains the `retry-after`
+      header, the value of the header is used as the next retry delay.
 
     * `:max_retries` - maximum number of retry attempts, defaults to `2` (for a total of `3`
       requests to the server, including the initial one.)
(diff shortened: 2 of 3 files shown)


test/req/steps_test.exs
```diff
@@ -698,6 +698,47 @@ defmodule Req.StepsTest do
     assert Agent.get(:counter, & &1) == 3
   end
 
+  @tag :capture_log
+  test "retry: retry-after", c do
```
wojtekmach, line 702: we could set retry_delay to a really high value and expect it not to be used, otherwise our test would time out. To keep the test super fast we can set retry-after to be 0:

```elixir
  @tag :capture_log
  test "retry: retry-after" do
    adapter = fn request ->
      response =
        case Map.get(request.private, :req_retry_count, 0) do
          0 -> Req.Response.new(status: 429) |> retry_after(0)
          1 -> Req.Response.new(status: 429) |> retry_after(DateTime.utc_now())
          2 -> Req.Response.new(status: 200, body: "ok")
        end

      {request, response}
    end

    assert Req.request!(adapter: adapter, retry_delay: 10000).body == "ok"
  end

  defp retry_after(r, value), do: Req.Response.put_header(r, "retry-after", retry_after(value))
  defp retry_after(integer) when is_integer(integer), do: to_string(integer)
  defp retry_after(%DateTime{} = dt), do: Req.Steps.format_http_datetime(dt)
```

we'd just need to make our slight change in production code, something like:

```diff
  delay
  |> ...
  |> DateTime.diff(DateTime.utc_now(), :millisecond)
+ |> max(0)
```

wdyt?

lib/req/steps.ex
```diff
@@ -1048,7 +1050,8 @@ defmodule Req.Steps do
           and returns boolean whether to retry
 
     * `:retry_delay` - sleep this number of milliseconds before making another attempt, defaults
-      to `2000`
+      to `#{@default_retry_delay}`. Will be ignored if HTTP 429 response includes `Retry-After`
+      header, in favor of the delay interval specified in the header.
```
wojtekmach, line 1054:
```suggestion
      to `#{@default_retry_delay}`. If response is HTTP 429 and contains `retry-after` header,
      the value of the header is used as the next retry delay.
```

lib/req/steps.ex
```diff
@@ -1132,6 +1135,32 @@ defmodule Req.Steps do
...
+    end
+  end
+
+  defp get_retry_delay(request, _response_or_exception) do
+    Map.get(request.options, :retry_delay, @default_retry_delay)
+  end
+
+  defp retry_delay_in_ms(delay_value) do
+    case Integer.parse(delay_value) do
+      {seconds, _} ->
```
wojtekmach, line 1154:
```suggestion
      {seconds, ""} ->
```

test/req/steps_test.exs
srowley: I like it! There are a few things I wonder about:

1. Depending on the internal details of retry counting
2. We don't really know anything about the delay that was/was not applied except by virtue of watching the test suite run and seeing that it wasn't really slow.

I'll work on my suggestion for addressing those two items.

wojtekmach: Both are valid concerns. Regarding 1, yeah, if internals change, the test will have to change too but I'm fine with that and fwiw the test will spectacularly fail so I'm not worried about not catching this. Regarding 2, besides noticing it is really slow, ExUnit will crash the test when it takes too long. The default timeout is 60s so my proposed delay was way too slow. We'd either do way bigger delay or `@tag timeout: 1000`.

wojtekmach: Oh, btw, the beauty (if I may say so) of using custom adapter is we could do this:

```elixir
    adapter = fn request ->
      attempt = Req.Request.get_private(request, :attempt, 0)

      response =
        case attempt do
          0 -> Req.Response.new(status: 429) |> retry_after(0)
          1 -> Req.Response.new(status: 429) |> retry_after(DateTime.utc_now())
          2 -> Req.Response.new(status: 200, body: "ok")
        end

      request = Req.Request.put_private(request, :attempt, attempt + 1)
      {request, response}
    end
```

and it would totally work too. Now, we'd still be coupled to how retry/1 works, that it re-uses request struct, but that's something way more stable. In fact if this sounds good to you why we don't ship with this? I guess We can skip send/assert_receive ping too, I don't think it adds much after all.

wojtekmach: I just added `update_private` so it can be even simpler:

```elixir
adapter = fn request ->
  request = Req.Request.update_private(request, :attempt, 0, & &1 + 1)
  attempt = request.private.attempt
  # ...
```

srowley:
Awesome - always more to learn - thank you!

CHANGELOG.md
```diff
@@ -310,8 +310,10 @@ See "Adapter" seciton in `Req.Request` module documentation for more information
   * `follow_redirects`: Don't send auth headers on redirect to different scheme/host/port
     unless `location_trusted: true` is set
 
+  * `retry`: The `:retry` option handles the `Retry-After` response header on HTTP 429 responses
```
wojtekmach, line 313:
```suggestion
  * `retry`: The `Retry-After` response header on HTTP 429 responses is now respected
```

wojtekmach:
Thank you!

PR #117, opened 2022-06-04
allow `:retry_delay` to be a `{&fun/1, initial}`

### LINKED ISSUE #115: `retry`: Allow `:retry_delay` to be a `{&fun/1, initial}`
this way a simple exponential backoff (1s, 2s, 4s, 8s, ...) could be defined as: 

```elixir
Req.new(retry_delay: {fn state -> {state, state * 2} end, 1000})
```

That backoff should be the new default too. And if someone wants an exponential backoff with jitter, it should be possible to do that too.

srowley:
Closes #115.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         include:
           - pair:
-              elixir: '1.11'
+              elixir: '1.12'
               otp: 24.3
           - pair:
               elixir: '1.13'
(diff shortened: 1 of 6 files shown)


wojtekmach:
@srowley Thank you!

I thought we'd need to keep some state around to implement jitter but looking at your example we don't, which has been a huge help. I mean, if someone needs that level of flexibility we'd maybe revisit, but I think this is totally good enough.

If we do use jitter, I noticed that the initial delay never goes through jitter which felt a little bit awkward.

I decided to change the contract from `{fn delay -> delay end, initial}` to: `fn retry_count -> delay end`. The default delay is: `&Integer.pow(2, &1) * 1000`. @aleDsz previously suggested to me to use retry count as the argument so thank you Ale for that foresight.

I decide to also soft-deprecate passing static delay. Instead of `retry_count: 1000` the user would do: `retry_count: fn _ -> 1000 end` which isn't too much hassle I think. But let's see how it goes in the wild before emitting a warning.

PR #119, opened 2022-06-15
`run_finch`: Add `:connect_options`

wojtekmach:


lib/req/application.ex
```diff
@@ -6,14 +6,21 @@ defmodule Req.Application do
...
-        id: Req.FinchHTTP1
-      ),
-      Supervisor.child_spec({Finch, name: Req.FinchHTTP2, pools: %{default: [protocol: :http2]}},
-        id: Req.FinchHTTP2
-      )
+      {DynamicSupervisor, name: Req.FinchSupervisor}
     ]
 
-    Supervisor.start_link(children, strategy: :one_for_one)
+    with {:ok, _sup} <- Supervisor.start_link(children, strategy: :one_for_one) do
```
jonatanklosko, line 12: We should return this at the end.

wojtekmach: oops, thank you!

wojtekmach:
OK, `Req.new(connect_timeout: timeout, cacertfile: cacertfile, ...)` it is. Except that we're also thinking about introducing a `:telemetry_metadata` field to Finch, to attach arbitrary metadata values to the emitted events. We'd need that in two places:

```elixir
Finch.start_link(name: MyFinch, telemetry_metadata: [foo: :bar])
Finch.request(request, MyFinch, telemetry_metadata: [foo: :bar])
```

and so in Req we can't have: `Req.new(telemetry_metaata: [foo: :bar])` because we don't know which one we refer to - with this PR we can now start dynamic pools on demand.

So maybe we have this after all:

```elixir
Req.new(
  connect_options: [
    timeout: 15000,
    cacertfile: cacertfile,
    telemetry_metadata: [service: :foo]
  ],
  request_options: [
    telemetry_metadata: [service: :foo]
  ]
)
```

setting these options is slightly less common so should be fine even if more verbose. wdyt @jonatanklosko @sneako @philss?

wojtekmach:
the idea was that yes we'd have: `request_options: [pool_timeout: ..., telemetry_metadata: ...]`. But I really don't like the name `request_options` as it's too vague and also we have: `request.options.request_options`. Following your suggestion I think we have:

```elixir
Req.new(
  connect_options: [
    timeout: 15000,
    cacertfile: cacertfile,
    telemetry_metadata: [service: :foo]
  ],
  telemetry_metadata: [service: :foo],
  receive_timeout: 5000
)
```

WDYT?

jonatanklosko:
Sounds good!

PR #131, opened 2022-08-23
Add none raiseing shorthands

torepettersen:
Hi and thanks for creating this great library.

Was just curious if it would make sense to add none raising shorthands for `get`, `post`, `put`, `patch`, `delete` and `head`?

I wanted that for myself, so I thought it would make sense to share :blush:


wojtekmach:
Thank you!

Issue #122, opened 2022-07-03
Issue with malformed `base_url`s


maxxst:
Hey everyone,

I just ran into an issue where the following code results in a 404 error:

```elixir
settings = Req.new(
  base_url: "https://api.apaleo.com/settings/v1",
  auth: {:bearer, access_token},
  headers: [
    "content-type": :"application/json",
  ]
)

Req.get!(settings, url: "/properties/TEST")
```

Thats because `URI.merge("https://api.apaleo.com/settings/v1", "/properties/TEST")` 
results in `"https://api.apaleo.com/properties/TEST"`.

I think it should be mentioned somewhere that a trailing `/` in the `:base_url` an **no** leading one in the `:url` are less error prone.

Maybe the readme should feature that style?

wojtekmach:
Thank you for the report. This is a bug, what you expected should work. A PR is appreciated.

wojtekmach:
btw,

```elixir
headers: [
  "content-type": :"application/json",
]
```

can be also written as:

```elixir
headers: [
  "content-type": "application/json",
]
```

or

```elixir
headers: [
  content_type: "application/json",
]
```

it is fine as is of course :)

maxxst:
But what is the intended way? Would a simple 

```elixir
"https://api.apaleo.com/settings/v1" <> "/properties/TEST"
```

more sufficient than 

```elixir
URI.merge("https://api.apaleo.com/settings/v1", "/properties/TEST")
```

otherwise we should define the behavior beforehand and than implement it.

wojtekmach:
not quite. If :url contains a full url (has schema), :base_url is ignored. That's why we went with URI.merge as it does it already.

wojtekmach:
in terms of specifying the behaviour I'm thinking about something like this. wdyt?

| `base_url`        |  `url`          | `result`            |
| ----------------- | --------------- | ------------------- |
| `https://a.com`   | `x`             | `https://a.com/x`   |
| `https://a.com`   | `/x`            | `https://a.com/x`   |
| `https://a.com/x` | `y`             | `https://a.com/x/y` |
| `https://a.com/x` | `/y`            | `https://a.com/x/y` |
| `https://a.com`   | `https://b.com` | `https://x`         |

but happy to make it stricter, reject some of the values, if that would make the implementation simpler and/or be more predictable.

maxxst:
| base_url           | url           | expected result   | URI result          | Comments           |
[...]

PR #132, opened 2022-08-24
`base_url` fix

torepettersen:
Addressing issue https://github.com/wojtekmach/req/issues/122

Stole idea from Tesla: https://github.com/elixir-tesla/tesla/blob/master/lib/tesla/middleware/core.ex#L33-L52

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -36,12 +36,29 @@ defmodule Req.Steps do
   def put_base_url(request)
 
   def put_base_url(%{options: %{base_url: base_url}} = request) do
-    url = URI.merge(base_url, request.url)
-    %{request | url: url}
+    url = to_string(request.url)
+
+    if Regex.match?(~r/^https?:\/\//i, url) do
+      # skip if url is already with scheme
+      request
+    else
+      %{request | url: URI.parse(join(base_url, url))}
+    end
   end
 
   def put_base_url(request) do
     request
+  end
+
+  defp join(base, url) do
+    case {String.last(base), url} do
+      {nil, url} -> url
+      {"/", "/" <> rest} -> base <> rest
+      {"/", rest} -> base <> rest
+      {_, ""} -> base
+      {_, "/" <> rest} -> base <> "/" <> rest
+      {_, rest} -> base <> "/" <> rest
+    end
   end
 
   @doc """
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -36,13 +36,11 @@ defmodule Req.Steps do
   def put_base_url(request)
 
   def put_base_url(%{options: %{base_url: base_url}} = request) do
-    url = to_string(request.url)
-
-    if Regex.match?(~r/^https?:\/\//i, url) do
+    if request.url.scheme != nil do
       # skip if url is already with scheme
       request
     else
-      %{request | url: URI.parse(join(base_url, url))}
+      %{request | url: URI.parse(join(base_url, request.url))}
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/req/steps.ex
```diff
@@ -36,14 +36,31 @@ defmodule Req.Steps do
   def put_base_url(request)
 
   def put_base_url(%{options: %{base_url: base_url}} = request) do
-    url = URI.merge(base_url, request.url)
-    %{request | url: url}
+    url = to_string(request.url)
+
+    if Regex.match?(~r/^https?:\/\//i, url) do
+      # skip if url is already with scheme
+      request
```
wojtekmach, line 43: Instead of matching the url string, would this work?

```elixir
if request.url.scheme != nil do
```

lib/req/steps.ex
```diff
@@ -36,14 +36,31 @@ defmodule Req.Steps do
...
+    end
   end
 
   def put_base_url(request) do
     request
   end
 
+  defp join(base, url) do
+    case {String.last(base), url} do
+      {nil, url} -> url
```
wojtekmach, line 55: is `base_url` being `nil` something that can occur in practice? If so, please add a test for this if one doesn't exist.

wojtekmach (commented):
sorry for late reply!

lib/req/steps.ex
torepettersen: Yes, that seems a lot better.

lib/req/steps.ex
```diff
@@ -36,14 +36,31 @@ defmodule Req.Steps do
...
+      %{request | url: URI.parse(join(base_url, url))}
+    end
   end
 
   def put_base_url(request) do
     request
   end
 
+  defp join(base, url) do
+    case {String.last(base), url} do
```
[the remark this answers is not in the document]
torepettersen: Can you elaborate? I don't think I understand what you want me to do.

lib/req/steps.ex
torepettersen: No, you are right. I can't come up with a use case where this would be helpful either.

lib/req/steps.ex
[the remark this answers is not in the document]
wojtekmach: Please change String.last to :binary.last :)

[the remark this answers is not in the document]
torepettersen: :+1:

lib/req/steps.ex
```diff
@@ -36,14 +36,28 @@ defmodule Req.Steps do
...
+    end
   end
 
   def put_base_url(request) do
     request
   end
 
+  defp join(base, url) do
+    case {:binary.last(base), to_string(url)} do
+      {"/", "/" <> rest} -> base <> rest
```
wojtekmach, line 53: we're missing some tests because this logic is wrong. Could you look into missing tests?

```
iex> :binary.last("foo/")
47
```

that's why I mentioned we need to match on `?/`, the `/` char code:

```
iex> ?/
47
```

torepettersen: Thanks, I was not aware that `?/` was the charcode for `/`. That was why that seems a bit cryptic to me.

But seems like the tests are failing because of connection issues. I am not able to reproduce locally.

Also am not able to understand how `test/req/steps_test.exs:1019`, was affected by this change. The request does not have a `base_url`, so might be a flaky one?

wojtekmach: Yes it's a flaky test, don't worry about it.

As I mentioned earlier, we had a wrong logic (`:binary.last` + `"/"`) that was not caught by any of the new tests so I'd appreciate it if you could investigate it further and either add more tests, change the logic, or bit of both.

torepettersen: Sorry, I feel so stupid for not seeing that. So thanks for keeping an eye on me.

Seems like Bypass accepts a path like `/api/v2//foo` when I specify a path like `/api/v2/foo`. But doing an assert that checks `conn.request_path` would have caught the bug I made with `:binary.last` + `"/"`.

PR #136, opened 2022-09-05
Add naive multipart/form-data capabilities to `Req.Steps.encode_body/1`.

Doerge:
Hi,

Thanks for a great package! I have simplified a couple of projects a lot with it! 🎉 

This PR adds multipart/form-data encoding for very simple forms. (I ran into an API that only accepts this).

Some notes:
- Files are not handled.
- I have a branch [here](https://github.com/Doerge/req/commit/13ea403e1d2cb43f9de54fd0d89d1c5ec254794b) with `Stream` body. Not sure if that is a good or bad idea in the overall architecture of Req, so I left it out for now. Maybe File streaming could be implemented on top of that?
- Tesla has a more solid implementation [here](https://github.com/elixir-tesla/tesla/blob/308a6761f012a21a01e68f0bf3e2dc409b44c7f9/lib/tesla/multipart.ex). Maybe someone wiser than me could cook up a better solution.

Let me know what you think of both this basic multipart/form-data PR, and the Stream branch.

Best,

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -100,6 +100,7 @@ defmodule Req do
           :params,
           :auth,
           :form,
+          :form_multi,
           :json,
           :compress_body,
           :compressed,
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -289,9 +289,47 @@ defmodule Req.Steps do
         |> Req.Request.put_new_header("content-type", "application/json")
         |> Req.Request.put_new_header("accept", "application/json")
 
+      data = request.options[:form_multi] ->
+        # Generate a random boundary
+        boundary =
+          :crypto.strong_rand_bytes(16)
+          |> Base.encode32(case: :lower, padding: false)
+
+        body =
+          encode_body_multipart_form(data, boundary)
+          |> Enum.to_list()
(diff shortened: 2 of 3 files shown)


wojtekmach:
Thank you for the PR. Before moving forward with this, I'd like to ask if you'd be interested in proposing API for a complete solution. There's no rush in implementing any of it but I'd like to get an idea to avoid a situation where we paint ourselves into the corner.

Regarding request body streaming, setting `body: {:stream, enum}`, I'd like to hold off on that unless you think it is essential for multipart.

Thank you!

Doerge:
Re. streams, I asked in the Elixir Slack channel and LostKobraKai was kind enough to answer. According to him there is no way to match on streams. Hence `{:stream, stream}` (or something similar) is necessary.

wojtekmach:
Yeah, _when_ we go with request streaming it will most likely be `{:stream, enumerable}` indeed as that's what Finch uses. :)

wojtekmach:
As mentioned earlier I'm very keen on a more comprehensive proposal before moving forward with any code changes so I'm going to close this for now. Thank you for starting the discussion and implementation, looking forward to continuing it!

Issue #139, opened 2022-09-20
Proposal: Allow all non-option fields in `Req.Request` to be specified as options


zachallaun:
Hi there! First off, thanks for your work on req. I've been using it as my primary request library and really enjoying it, but wanted to offer a proposal that I think would alleviate a current source of friction.

**Proposal:**

All non-options `Req.Request` fields (`:adapter, :url, :method, etc.`) passed to `Req.new/1` should be able to be merged into the request as options. I think the cleanest way to do this is to move much of the code in `Req.new/1` to `Request.merge_options/2`, and to use it in the implementation of `Req.new`.

**Context:**

I've been really enjoying using `Req` for some API work I'm doing. I'm interacting with multiple APIs, both REST and GraphQL, and have been using `Req` for the whole thing. My strategy has been to define a general `APIs.Helpers.base_req/1` that sets up general-purpose, shared logging/etc. and then `APIs.SomeAPI.base_req/1` that wraps it and sets up API-specific stuff like the authentication mechanism, etc.

This has worked really well and been really flexible and easy to work with, except for one point of friction: the use of `Req.Request` fields vs. options. There were a number of cases where I was using something like `Request.merge_options` when setting up a request and then getting strange results, only to find that I was actually merging in something that is special-cased in `Req.new` and attached to the request directly. Ultimately, you have to look at the type spec to determine what you need to do -- whether something needs to be `Map.put` directly into the request, or merged in as an option. 

This ends up making composition more difficult. For instance, if I pass `:url` in to `Req.new/1`, it will parse the URI for me. But if I merge it in later, I need to remember to parse the URI myself. This actually just bit me [here](https://github.com/wojtekmach/req/blob/main/lib/req.ex#L54) -- prior to v3.1, you could merge in a `:url` string (even though it violates the typespec) so long as you were using a `:base_url`, because `join/2` allowed the URL to be a string instead of a `URI`. Now obviously this was a bug in my code -- the typespec said it should have been a `URI.t`, but I didn't check and it worked. I think it highlights the problem, however. It would be better if I could `Request.merge_options(req, url: "/some_path")` and not have to concern myself with it.

_Note:_ Digging through the code while writing this, I realized that the current behavior may actually be considered a bug, since [`validate_options/2`](https://github.com/wojtekmach/req/blob/12a0c000265c7542345407c69f02a1b474d746af/lib/req/request.ex#L712) allows for `:method`, `:url`, etc.

wojtekmach:
[...]

PR #142, opened 2022-10-06
Add test conveniences

mcrumm:
Hi @wojtekmach– thanks for making Req! :)

You did not ask for this, so if you are not interested in it please feel free to close this PR. After using Req for a while, especially while building [AbsintheClient](https://github.com/CargoSense/absinthe_client), I found myself several times reaching for a tool to introspect requests to ensure that steps were being applied how I expected. I was talking with @elbow-jason and he had the idea for a "dry run" function that could provide the request after all the steps had occurred, so that's what we have here. It can override the response for testing, too. Wdyt?

### DIFF
--- a/lib/req/test.ex
+++ b/lib/req/test.ex
@@ -1,0 +1,171 @@
+defmodule Req.Test do
+  @moduledoc """
+  Conveniences for testing requests and responses.
+
+  This module is useful mostly for testing steps and plugins. It provides
+  functions to preview a request after all of its steps have been executed.
+  You can also test the result of the response and error pipelines by
+  overriding the response or exception returned by the test adapter.
+  """
+  alias Req.{Request, Response}
+
+  @doc """
+  Performs a test run on the request.
+
+  Refer to `run/2` for more information.
+
+  ## Examples
+
+      iex> {:ok, res} = Req.new(base_url: "http://httpbin.org") |> Req.Test.run()
+      iex> res.body.url
+      URI.parse("http://httpbin.org")
+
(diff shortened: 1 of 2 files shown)


wojtekmach:
There is a Req.Request.prepare/1 that does something very similar. There were some issues with it so I `@doc false` it pending removal, but I don’t remember at the moment what they were so I will get back to you. In any case, I’ll review the PR soon after.

wojtekmach:
btw, in case you weren't aware, I was hoping passing a custom adapter will solve a lot of scenarios in tests. An adapter is just a request step, the only special thing it is always the _last_ one. Here's an example where we can both inspect the request as well as return a canned response:

```elixir
fake = fn request ->
  {dbg(request), Req.Response.json(%{x: 1})}
end
%{"x" => 1} = Req.request!(adapter: fake).body
```

mcrumm:
Absolutely! This PR is mostly just a long-winded way of formalizing:

```elixir
Req.request!(adapter: &{&1, Response.new(body: &1)}).body
```

...as a way to test, and now that I have written it that succinctly I think we do not need a whole module of functions for it :)

Fwiw my desire for something like this arose after trying to set some request keys like `:headers` through `Req.Request.merge_options/2` only to realize it didn't work when my response had errors because my request headers were not set. Addressing #139 would eliminate much of the uncertainty.

wojtekmach:
> Addressing https://github.com/wojtekmach/req/issues/139 would eliminate much of the uncertainty.

sounds good, let's focus on #139 then. I'm going to close this for now as I hope some enhancements to the current API should be enough to cover this but do let me know otherwise. Thanks!

PR #145, opened 2022-10-16
Decode JSON when response is json-api mime type

rodrigues:
Hi Wojtek! 👋

I'm not sure whether this would be the path to go, but just noticed that `json-api` responses don't call jason decode step because of the mime type, WDYT?

Cheers!

### DIFF
--- a/test/req/steps_test.exs
+++ b/test/req/steps_test.exs
@@ -320,6 +320,16 @@ defmodule Req.StepsTest do
     assert Req.get!(c.url).body == %{"a" => 1}
   end
 
+  test "decode_body/1: json-api", c do
+    Bypass.expect(c.bypass, "GET", "/", fn conn ->
+      conn
+      |> Plug.Conn.put_resp_header("content-type", "application/vnd.api+json; charset=utf-8")
+      |> json(200, %{a: 1})
+    end)
+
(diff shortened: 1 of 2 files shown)


test/req/steps_test.exs
```diff
@@ -320,6 +320,16 @@ defmodule Req.StepsTest do
     assert Req.get!(c.url).body == %{"a" => 1}
   end
 
+  test "decode_body/1: json-api", c do
+    Bypass.expect(c.bypass, "GET", "/", fn conn ->
+      conn
+      |> Plug.Conn.put_req_header("content-type", "application/vnd.api+json; charset=utf-8")
+      |> json(200, %{a: 1})
```
wojtekmach, line 327: this test passes on the main branch already but for the wrong reason. It is because the `json/3` function incorrectly set the `application/json` content-type whereas it should only do that if content-type is not set on the request. Could you fix that one while at it too?

rodrigues: updated `json/3` to not override the content-type, also confirmed this test breaks in main, thanks!

wojtekmach:
Ohhh, the test was using the private json/3, I thought it was using Req.Response.json/3. 😅 if it would need a similar change and you’d like to send a PR it would be very appreciated but there’s of course no rush. Thanks!

PR #146, opened 2022-10-16
Adapt Req.Response.json/3 to not override content type

rodrigues:
Follow up on https://github.com/wojtekmach/req/pull/145#issuecomment-1279993636

PS: I've sent you a DM on twitter  👍

### DIFF
--- a/lib/req/response.ex
+++ b/lib/req/response.ex
@@ -54,12 +54,28 @@ defmodule Req.Response do
         headers: [{"content-type", "application/json"}],
         body: ~s|{"hello":42}|
       }
+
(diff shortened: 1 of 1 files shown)


lib/req/response.ex
```diff
@@ -54,12 +54,28 @@ defmodule Req.Response do
         headers: [{"content-type", "application/json"}],
         body: ~s|{"hello":42}|
       }
+
```
wojtekmach, line 57:
```suggestion

If the request already contains a 'content-type' header, it is kept as is.

```

wojtekmach:
Thank you!

PR #149, opened 2022-11-07
Add proxy to connection options

un3481:
Hello,

I am using your package for a project of mine and it works very well. But at some point I had to use a proxy and I realized that I could not pass a `:proxy` option directly to the request function call.

I really needed this feature, so I added a `:proxy` option that the user can pass along with `:timeout` and `:protocol`. The proxy must be provided in a format supported by the [Mint](https://github.com/elixir-mint/mint) package: A tuple `{scheme, address, port, opts}` that identifies the proxy to connect to. 

```elixir
# proxy with scheme 'https', ip '34.91.135.38' and port '80'
https_proxy = {:https, "34.91.135.38", 80, []}

req = Req.get!("https://api.github.com/repos/sneako/finch", proxy: https_proxy)
```
Also note that I did not implement any extra tests after I added this feature. That's because I assume this is not a significant change for this project as I'm just passing an extra option to [Finch](https://github.com/sneako/finch) package.

I am already using this PR as a dependency in a personal project of mine called [exgooglenews](https://github.com/un3481/exgooglenews) and the proxy feature works just like I intended it to. What I want now is to have this feature in your Hex package so I can use your official package as a dependency.

Let me know what you think of this change.

Best regards.

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -700,6 +700,8 @@ defmodule Req do
 
         * `:protocol` - the HTTP protocol to use, defaults to `:http1`.
 
+        * `:proxy` - a proxy to use given in a format supported by the Mint package: A tuple `{schema, address, port, options}`.
+
     * `:pool_timeout` - pool checkout timeout in milliseconds, defaults to `5000`.
 
     * `:receive_timeout` - socket receive timeout in milliseconds, defaults to `15_000`.
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -700,7 +700,9 @@ defmodule Req do
 
         * `:protocol` - the HTTP protocol to use, defaults to `:http1`.
 
-        * `:proxy` - a proxy to use given in a format supported by the Mint package: A tuple `{schema, address, port, options}`.
+        * `:proxy` - a `{schema, address, port, options}` tuple. See
+          [Mint "Proxying" documentation](https://hexdocs.pm/mint/Mint.HTTP.html#connect/4-proxying)
+          for more information.
 
     * `:pool_timeout` - pool checkout timeout in milliseconds, defaults to `5000`.
 
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -475,7 +475,9 @@ defmodule Req.Steps do
 
         * `:protocol` - the HTTP protocol to use, defaults to `:http1`.
 
-        * `:proxy` - a proxy to use given in a format supported by the Mint package: A tuple `{schema, address, port, options}`.
+        * `:proxy` - a `{schema, address, port, options}` tuple. See
+          [Mint "Proxying" documentation](https://hexdocs.pm/mint/Mint.HTTP.html#connect/4-proxying)
+          for more information.
 
     * `:pool_timeout` - pool checkout timeout in milliseconds, defaults to `5000`.
 
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -512,10 +512,15 @@ defmodule Req.Steps do
         :error ->
           cond do
             options = request.options[:connect_options] ->
-              Req.Request.validate_options(options, MapSet.new([:timeout, :protocol]))
+              Req.Request.validate_options(options, MapSet.new([:timeout, :proxy, :protocol]))
 
               pool_opts = [
-                conn_opts: [transport_opts: [timeout: options[:timeout] || 30_000]],
+                conn_opts: [
+                  proxy: options[:proxy] || nil,
+                  transport_opts: [
+                    timeout: options[:timeout] || 30_000
+                  ]
+                ],
                 protocol: options[:protocol] || :http1
               ]
 
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -475,6 +475,8 @@ defmodule Req.Steps do
 
         * `:protocol` - the HTTP protocol to use, defaults to `:http1`.
 
+        * `:proxy` - a proxy to use given in a format supported by the Mint package: A tuple `{schema, address, port, options}`.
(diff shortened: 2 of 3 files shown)


wojtekmach (commented):
Thank you for the PR! Agreed that we're just passing options down to Finch but let's add a test anyway to make sure it doesn't accidentally break in the future. Please also document this option  in Req.request/1 and Req.Steps.run_finch/2.

lib/req/steps.ex
```diff
@@ -475,6 +475,8 @@ defmodule Req.Steps do
 
         * `:protocol` - the HTTP protocol to use, defaults to `:http1`.
 
+        * `:proxy` - a proxy to use given in a format supported by the Mint package: A tuple `{schema, address, port, options}`.
```
wojtekmach, line 478:
```suggestion
        * `:proxy` - a `{schema, address, port, options}` tuple. See
          [Mint "Proxying" documentation](https://hexdocs.pm/mint/Mint.HTTP.html#connect/4-proxying)
          for more information.
```

mix.exs
```diff
@@ -1,7 +1,7 @@
 defmodule Req.MixProject do
   use Mix.Project
 
-  @version "0.3.1"
+  @version "0.3.2"
```
wojtekmach, line 4: Let's not change versions (or changelogs etc) on PRs as we don't know when we'll release it, maybe it will be 0.3 or maybe 0.4.

```suggestion
  @version "0.3.1"
```

lib/req/steps.ex
```diff
@@ -512,10 +514,15 @@ defmodule Req.Steps do
...
-              Req.Request.validate_options(options, MapSet.new([:timeout, :protocol]))
+              Req.Request.validate_options(options, MapSet.new([:timeout, :protocol, :proxy]))
 
               pool_opts = [
-                conn_opts: [transport_opts: [timeout: options[:timeout] || 30_000]],
+                conn_opts: [
+                  transport_opts: [
+                    timeout: options[:timeout] || 30_000
+                  ],
+                  proxy: options[:proxy]
```
wojtekmach, line 524: does Mint accept `proxy: nil` or we need to conditionally set the proxy key? Maybe that's why we have test failures?

un3481:
Hello,

I added a small documentation about the :proxy option, but also referenced that this option is passed to Finch (and Mint).

I also implemented a single test for the proxy feature and it works. I'm using Bypass, just like all the other tests, so no need to add extra packages. 

But while testing I found out that one other test from your package had a problem, and it's not related to anything I did, it was there before. The error I got was this:

```
..........................................................

  1) test run_finch/1: :connect_options :timeout (Req.StepsTest)
     test/req/steps_test.exs:1012
     ** (CaseClauseError) no case clause matching: {:ok, nil}
     code: assert Req.request(req) == {:error, %Mint.TransportError{reason: :timeout}}
     stacktrace:
       (mint 1.4.2) lib/mint/http.ex:359: Mint.HTTP.connect/4
       (finch 0.13.0) lib/finch/http1/conn.ex:45: Finch.Conn.connect/1
       (finch 0.13.0) lib/finch/http1/pool.ex:44: anonymous fn/8 in Finch.HTTP1.Pool.request/5
       (nimble_pool 0.2.6) lib/nimble_pool.ex:349: NimblePool.checkout!/4
       (finch 0.13.0) lib/finch/http1/pool.ex:38: Finch.HTTP1.Pool.request/5
       (finch 0.13.0) lib/finch.ex:324: anonymous fn/4 in Finch.request/3
       (telemetry 1.1.0) /home/helvetia/Documentos/GitHub/req/deps/telemetry/src/telemetry.erl:320: :telemetry.span/3
       (req 0.3.2) lib/req/steps.ex:554: Req.Steps.run_finch/1
       (req 0.3.2) lib/req/request.ex:634: Req.Request.run_request/2
       test/req/steps_test.exs:1023: (test)

.........................
Finished in 4.6 seconds (4.6s async, 0.00s sync)
47 doctests, 74 tests, 1 failure, 37 excluded

```

This error happens during test `"run_finch/1: :connect_options :timeout"` in "steps_test.exs" file at line 1012. I don't know exactly what must be done to fix this test, but I could try to help.

Let me know what you think.

Best Regards.

lib/req/steps.ex
wojtekmach: `proxy: nil` seems not to be ok:

```
iex> Mix.install([:mint])
iex> Mint.HTTP.connect(:http, "httpbin.org", 80)
{:ok, %Mint.HTTP1{}}
iex> Mint.HTTP.connect(:http, "httpbin.org", 80, proxy: nil)
** (CaseClauseError) no case clause matching: {:ok, nil}
    (mint 1.4.2) lib/mint/http.ex:359: Mint.HTTP.connect/4
    iex:5: (file)
```

lib/req.ex
```diff
@@ -700,6 +700,8 @@ defmodule Req do
 
         * `:protocol` - the HTTP protocol to use, defaults to `:http1`.
 
+        * `:proxy` - a proxy to use given in a format supported by the Mint package: A tuple `{schema, address, port, options}`.
```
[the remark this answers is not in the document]
wojtekmach: could you re-apply this (and the other one) suggestion if you don't mind? It seems it got overriden somehow.

lib/req/steps.ex
un3481: Hello, after testing again I figured out that the error in `:timeout` test was indeed caused by `proxy: nil`.
I can't explain why only this test had this issue and why the request appears to work with `proxy: nil`, but I removed this possibility and now all tests are passing.

wojtekmach:
Thank you!

PR #150, opened 2022-11-09
Don't repeat query string on retry

### LINKED ISSUE #140: URI parameters appear to be duplicated on retry
Given this script:

```elixir
Mix.install([
  {:req, "0.3.1"}
])

require Logger

Req.new(url: "https://httpbin.org/status/408", params: [test: "a"])
|> Req.Request.append_request_steps(
  log_request: fn request ->
    Logger.debug("#{request.method} #{URI.to_string(request.url)}")

    request
  end
)
|> Req.get!()
```

… I get this output:

```
11:09:15.145 [debug] get https://httpbin.org/status/408?test=a

11:09:15.692 [error] retry: got response with status 408, will retry in 1000ms, 3 attempts left

11:09:16.694 [debug] get https://httpbin.org/status/408?test=a&test=a

11:09:16.778 [error] retry: got response with status 408, will retry in 2000ms, 2 attempts left

11:09:18.780 [debug] get https://httpbin.org/status/408?test=a&test=a&test=a

11:09:18.868 [error] retry: got response with status 408, will retry in 4000ms, 1 attempt left

11:09:22.871 [debug] get https://httpbin.org/status/408?test=a&test=a&test=a&test=a
```

Notice how on the first request, I have a single `test=a` URI parameter, but on each subsequent attempt the URI parameters are duplicated.

Please let me know if I've done something wrongly here!

aleDsz:
- Closes #140

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -330,12 +330,16 @@ defmodule Req.Steps do
   end
 
   defp put_params(request, params) do
-    encoded = URI.encode_query(params)
+    if Req.Request.get_private(request, :req_retry_count) do
+      request
+    else
+      encoded = URI.encode_query(params)
 
-    update_in(request.url.query, fn
-      nil -> encoded
-      query -> query <> "&" <> encoded
-    end)
+      update_in(request.url.query, fn
+        nil -> encoded
+        query -> query <> "&" <> encoded
+      end)
+    end
   end
 
   @doc """
--- a/test/req/steps_test.exs
+++ b/test/req/steps_test.exs
@@ -835,6 +835,38 @@ defmodule Req.StepsTest do
     refute_received _
   end
 
+  @tag :capture_log
+  test "retry: don't repeat query string", c do
+    pid = self()
+
+    Bypass.expect(c.bypass, "GET", "/", fn conn ->
+      send(pid, :ping)
+      Plug.Conn.send_resp(conn, 500, "oops")
+    end)
+
+    request =
+      Req.new(url: c.url, params: [foo: "bar"], retry_delay: 1)
+      |> Req.Request.append_request_steps(
+        logger_request: fn request ->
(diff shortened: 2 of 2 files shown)


lib/req/steps.ex
```diff
@@ -330,12 +330,16 @@ defmodule Req.Steps do
   end
 
   defp put_params(request, params) do
-    encoded = URI.encode_query(params)
+    if Req.Request.get_private(request, :req_retry_count) do
+      request
```
wojtekmach, line 334: This is a pretty sensible way to fix the issue but I can't help but wonder if there's another way. There very well may not be a better option. But it means that any step that changes the request somehow, which is most of them, would need to have similar conditional and it is weird to hardcode it for retry.

Another example where this is clearly broken is compress_body/1 step, on retries we'll keep re-compressing the request body, getting a different result each time.

I think we have these options:

1. This PR
2. Require steps to be idempotent. For example, put_params/1 could check if given params are already in the query string. (Which would prevent us from sending duplicate ones.) compress_body/1 could check if content-encoding is already set and if so, it's a no-op, etc.
3. When a step is executed, remove the option that triggered it. This seems fragile.
4. When a step is executed, remove it from the steps list. I vaguely remember we had plugins which checked if a given request contains given steps, so this would break that.
5. Option 4 except we keep two lists, the latter being private and the one that gets stuff deleted from.

Any other ideas?

jonatanklosko: 4/5. when we run a step that returns on error, it is executed, but at the same time, for retry we would like to repeat that step, how do we distinguish that?

One other option would be to store the initial `%Req.Request{}` in a separate field when calling `run/1`, then if we need to re-run we just use that intact request, only bumping the private field. Though with this approach we redo stuff like compression.

jonatanklosko:
@wojtekmach sounds good to me!

PR #154, opened 2022-11-29
Inherit scheme from previous location on redirection if needed

lud:
This change allows to follow redirect locations such as `//other-host.com/a/b?c=d` where the scheme should be derived from the previous request.

I had to change the order of operations in `build_redirect_request/2` because I assumed that we want to put the scheme that will be used before comparing with the previous one in `remove_credentials_if_untrusted/2`.

Please let me know what you think.

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -1041,7 +1041,8 @@ defmodule Req.Steps do
     Logger.debug(["follow_redirects: redirecting to ", location])
 
     location_trusted = Map.get(request.options, :location_trusted)
-    location_url = URI.parse(location)
+
+    location_url = location |> URI.parse() |> inherit_from_url(request.url)
 
     request
     |> remove_params()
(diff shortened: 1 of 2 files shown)


wojtekmach:
Thank you!

lud:
Happy to participate!

lib/req/steps.ex
```diff
@@ -1041,7 +1041,8 @@ defmodule Req.Steps do
     Logger.debug(["follow_redirects: redirecting to ", location])
 
     location_trusted = Map.get(request.options, :location_trusted)
-    location_url = URI.parse(location)
+
+    location_url = location |> URI.parse() |> inherit_from_url(request.url)
```
wojtekmach, line 1045: I wonder if it could have been simply:

```elixir
location_url = URI.merge(request.url, URI.parse(location))
```

with limited testing it seems to do what we want:

```elixir
iex> test = fn a, b -> URI.merge(URI.parse(a), URI.parse(b)) |> URI.to_string() end

iex> test.("https://foo/foo?foo=1", "https://bar/bar?bar=1")
"https://bar/bar?bar=1"

iex> test.("https://foo/foo?foo=1", "//bar/bar?bar=1")
"https://bar/bar?bar=1"

iex> test.("https://foo/foo?foo=1", "/bar?bar=1")
"https://foo/bar?bar=1"
```

if that sounds good, a PR would be appreciated as always. :D

PR #155, opened 2022-11-30
use URI.merge to derive the redirect URL from the current URL

lud:
Following the conversation here : https://github.com/wojtekmach/req/pull/154

It seems that the `URI.merge/2` algorithm just does what we want.

(Reminder: this was to support urls such as `//somehost/path/to` when following redirects)


wojtekmach:
Thanks!

PR #157, opened 2022-12-07
`run_finch`: Add `:finch_request` option

aselder:
This PR allows an optional callback function to be passed as an options which allow modification of the Finch request object before it is executed.

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -80,11 +80,11 @@ defmodule Req do
           :cache_dir,
           :plug,
           :finch,
+          :finch_request,
           :connect_options,
           :receive_timeout,
           :pool_timeout,
-          :unix_socket,
-          :finch_request
+          :unix_socket
         ])
     }
     |> update(options)
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -84,7 +84,7 @@ defmodule Req do
           :receive_timeout,
           :pool_timeout,
           :unix_socket,
-          :customize_finch
+          :finch_request
         ])
     }
     |> update(options)
@@ -764,8 +764,8 @@ defmodule Req do
 
     * `:unix_socket` - if set, connect through the given UNIX domain socket
     
-    * `:customize_finch` - a function to modify the built Finch request before execution. This function takes a 
-       Finch request and returns a Finch request. Defaults to the identity function.
+    * `:finch_request` - a function to modify the built Finch request before execution. This function takes a 
+       Finch request and returns a Finch request. If not provided, the finch request will not be modified
 
   ## Examples
 
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -83,7 +83,8 @@ defmodule Req do
           :connect_options,
           :receive_timeout,
           :pool_timeout,
-          :unix_socket
+          :unix_socket,
+          :customize_finch
         ])
     }
     |> update(options)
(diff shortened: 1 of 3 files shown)


aselder:
This is a work in progress to show the path I'm proposing. If it's decided that this is a good path, I'll flesh it out some more, add tests, etc.

wojtekmach:
I like this change. With this we can trivially see the exact Finch request that is about to be sent:

```elixir
iex> Req.get!("https://api.github.com/repos/wojtekmach/req", customize_finch: &dbg/1).body["description"]
[iex:1: (file)]
x1 #=> %Finch.Request{
  scheme: :https,
  host: "api.github.com",
  port: 443,
  method: "GET",
  path: "/repos/wojtekmach/req",
  headers: [
    {"user-agent", "req/0.3.2"},
    {"accept-encoding", "zstd, br, gzip, deflate"}
  ],
  body: nil,
  query: nil,
  unix_socket: nil,
  private: %{}
}

"Req is a batteries-included HTTP client for Elixir."
```

and that would have been so convenient many times.

I'm not sure about `:customize_finch` for option name. How about `:finch_request`?

cc @sneako @jonatanklosko

jonatanklosko:
Sounds good to me!

sneako:
Nice, yeah I will definitely use this :)

lib/req.ex
```diff
@@ -762,6 +763,9 @@ defmodule Req do
     * `:receive_timeout` - socket receive timeout in milliseconds, defaults to `15_000`.
 
     * `:unix_socket` - if set, connect through the given UNIX domain socket
+    
+    * `:customize_finch` - a function to modify the built Finch request before execution. This function takes a 
+       Finch request and returns a Finch request. Defaults to the identity function.
```
jonatanklosko, line 768: We also need to document this option under `run_finch` :)

lib/req/steps.ex
```diff
@@ -517,10 +517,12 @@ defmodule Req.Steps do
   @doc step: :request
   def run_finch(request) do
     finch_name = finch_name(request)
+    customize_function = customize_function(request)
 
     finch_request =
       Finch.build(request.method, request.url, request.headers, request.body)
       |> Map.replace!(:unix_socket, request.options[:unix_socket])
+      |> customize_function.()
```
wojtekmach, line 525: wdyt about instead a private function that does something like this?

```elixir
defp update_finch_request(finch_request, request) do
  case Map.fetch(request.options, :finch_request) do
    {:ok, fun} -> fun.(finch_request)
    :error -> finch_request
  end
end
```

lib/req.ex
aselder: Yep, I'll finish up the docs and tests now that we like the general approach

wojtekmach:
Thank you!

wojtekmach:
@aselder are you still using custom `:finch_request` or by any chance you moved to using `opentelemetry_req`?

The next release will have [response body streaming](https://github.com/wojtekmach/req/blob/main/CHANGELOG.md#add-response-body-streaming) which couldn't be used with a custom :finch_request. Of course I could just error under that scenario. But while it is a very useful escape hatch, if you're not using it I'd rather deprecate it and remove on Req 1.0, to keep things simpler internally.

aselder:
Go ahead and release.. worst case we just stay pinned to the current version until we can get caught up... Don't hold a release on us.

wojtekmach:
Fwiw I _think_ we could have a callback that would do the Finch request but instead of abstracting over Finch.request we’d abstract over Finch.stream, or the new .stream_while ideally, and build everything on top of that.

wojtekmach:
Oh and just to be clear :finch_request as is will be supported until 1.0, you just wouldn’t be able to use it with the new response body streaming until we figure out a path.

PR #163, opened 2022-12-27
Allow customization of log level for retry messages

aselder:
Also, allows `false` to disable logging of retry messages.

### DIFF
--- a/test/req/steps_test.exs
+++ b/test/req/steps_test.exs
@@ -826,8 +826,7 @@ defmodule Req.StepsTest do
     request = Req.new(adapter: adapter, url: c.url, retry_delay: 1)
     log = ExUnit.CaptureLog.capture_log(fn -> Req.get!(request) end)
 
-    assert log =~
-             "[error] retry: got response with status 500, will retry in 1ms, 3 attempts lef"
+    assert log =~ "[error]"
   end
 
   @tag :capture_log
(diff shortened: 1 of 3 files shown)


test/req/steps_test.exs
```diff
@@ -805,6 +805,76 @@ defmodule Req.StepsTest do
...
+            Req.Response.new(status: 200, body: "ok")
+        end
+
+      {request, response}
+    end
+
+    request = Req.new(adapter: adapter, url: c.url, retry_delay: 1)
+    log = ExUnit.CaptureLog.capture_log(fn -> Req.get!(request) end)
+
+    assert log =~ "[error]"
```
wojtekmach, line 829: We had this failure:

```
code:  assert log =~ "[info] retry: got response with status 500, will retry in 1ms, 3 attempts lef"
     left:  "\n22:41:35.585 [info]  retry: got response with status 500, will retry in 1ms, 3 attempts left\n"
     right: "[info] retry: got response with status 500, will retry in 1ms, 3 attempts lef"
     stacktrace:
       test/req/steps_test.exs:853: (test)
```

In short, `[info] retry` vs `[info]  retry:`. No idea where the extra space came from. Do you know? I'd rather keep these tests a bit stricter given we run async: true and thus they may accidentally pass if some other test just logs the log tag.

aselder: I have no idea where the extra space came from. The tests were passing on my machine.. Let me look into getting it working with a regex.

aselder:
@wojtekmach Can you re-run the tests? Looks like the GitHub API flaked during the tests.

wojtekmach:
Thank you!

PR #171, opened 2023-03-01
Fix "(ArgumentError) unknown option :hostname"

hdahlheim:
According to the documentation it should be possible to pass a `:hostname` option to finch as part of `:connect_options`.

Passing the `:hostname` option resulted in an ArgumentError because `:hostname` was not part of the valid options MapSet.

This commit adds `:hostname` to the valid options MapSet.

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -611,7 +611,8 @@ defmodule Req.Steps do
                 :transport_opts,
                 :proxy_headers,
                 :proxy,
-                :client_settings
+                :client_settings,
+                :hostname
               ])
             )
 
--- a/test/req/steps_test.exs
+++ b/test/req/steps_test.exs
@@ -1203,6 +1203,25 @@ defmodule Req.StepsTest do
     assert Req.request!(req, url: "/foo/bar").body == "ok"
   end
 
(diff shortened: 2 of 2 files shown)


lib/req/steps.ex
```diff
@@ -611,7 +611,8 @@ defmodule Req.Steps do
                 :transport_opts,
                 :proxy_headers,
                 :proxy,
-                :client_settings
+                :client_settings,
+                :hostname
```
wojtekmach, line 615: Thank you for the PR. Note we're explicitly retrieving :hostname option elsewhere:

https://github.com/wojtekmach/req/blob/v0.3.5/lib/req/steps.ex#L618

Given your patch, that code was never correct. Can you remove it? Also, could you write a test for this change?

hdahlheim: If I'm not mistaken the explicit retrieving of `:hostname` still needs to happen for the option to work. 

The problem I faced was that passing `:hostname` raises on https://github.com/wojtekmach/req/blob/v0.3.5/lib/req/steps.ex#L606 because `:hostname` was not part of the MapSet.

wojtekmach:
Thank you!

PR #174, opened 2023-03-31
Update deps

rlopzc:
Finch now allows the next version of nimble_options.

Updating this dep lets OpenTelemetry Phoenix use a different version of nimble_options


rlopzc:
Bump @wojtekmach . What else can I do to help?

wojtekmach:
Thank you!

> Updating this dep lets OpenTelemetry Phoenix use a different version of nimble_options

mix.lock is local to req, it is not taken into consideration when req is used as a dependency. But your change is useful to req developers for sure. Thanks!

PR #177, opened 2023-04-18
Avoid assuming NaiveDateTime is GMT

lau:
Hi,

This is a nice library, thank you.

Here's a PR with the intent of avoiding bugs: Removing the automatic upgrading of `NaiveDateTime` to GMT.

In this case a `NaiveDateTime` is silently "upgraded" to be GMT even though we have no idea if that's true or not.

If a developer has a `NaiveDateTime` that the developer is sure is UTC, it is trivial to call `DateTime.from_naive!(naive, "Etc/UTC")`. However libraries shouldn't assume the UTC offset since it can lead to bugs. The reason that the `NaiveDateTime` type exists is for when we don't know the timezone. Libraries should not assume the time zone offset of `NaiveDateTime`.

This is consistent with the standard library where if you call `DateTime.from_iso8601("2023-04-18 06:16:28.609427")` you get an error because the offset is missing from the string. And why `DateTime.to_unix/1` does not accept `NaiveDateTime`. The `DateTime.to_unix/1` function is in `DateTime` rather than `NaiveDateTime` because unix timestamps are UTC by definition and `NaiveDateTime` does not contain UTC offsets. The same goes for HTTP datetimes.

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -460,8 +460,8 @@ defmodule Req.Steps do
   defp put_if_modified_since(request, cache_path) do
     case File.stat(cache_path) do
       {:ok, stat} ->
-        datetime = stat.mtime |> NaiveDateTime.from_erl!() |> format_http_datetime()
-        Req.Request.put_new_header(request, "if-modified-since", datetime)
+        http_datetime_string = stat.mtime |> NaiveDateTime.from_erl!() |> DateTime.from_naive!("Etc/UTC") |> format_http_datetime()
+        Req.Request.put_new_header(request, "if-modified-since", http_datetime_string)
 
       _ ->
         request
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -460,7 +460,11 @@ defmodule Req.Steps do
   defp put_if_modified_since(request, cache_path) do
     case File.stat(cache_path) do
       {:ok, stat} ->
-        http_datetime_string = stat.mtime |> NaiveDateTime.from_erl!() |> DateTime.from_naive!("Etc/UTC") |> format_http_datetime()
+        http_datetime_string =
+          stat.mtime
+          |> NaiveDateTime.from_erl!()
+          |> DateTime.from_naive!("Etc/UTC")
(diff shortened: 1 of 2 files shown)


lib/req/steps.ex
```diff
@@ -460,8 +460,8 @@ defmodule Req.Steps do
   defp put_if_modified_since(request, cache_path) do
     case File.stat(cache_path) do
       {:ok, stat} ->
-        datetime = stat.mtime |> NaiveDateTime.from_erl!() |> format_http_datetime()
-        Req.Request.put_new_header(request, "if-modified-since", datetime)
+        http_datetime_string = stat.mtime |> NaiveDateTime.from_erl!() |> DateTime.from_naive!("Etc/UTC") |> format_http_datetime()
```
wojtekmach, line 463:
```suggestion
        http_datetime_string =
          stat.mtime
          |> NaiveDateTime.from_erl!()
          |> DateTime.from_naive!("Etc/UTC")
          |> format_http_datetime()
```

lib/req/steps.ex
```diff
@@ -460,8 +460,12 @@ defmodule Req.Steps do
...
     case File.stat(cache_path) do
       {:ok, stat} ->
-        datetime = stat.mtime |> NaiveDateTime.from_erl!() |> format_http_datetime()
-        Req.Request.put_new_header(request, "if-modified-since", datetime)
+        http_datetime_string =
+          stat.mtime
+          |> NaiveDateTime.from_erl!()
+          |> DateTime.from_naive!("Etc/UTC")
+          |> format_http_datetime()
+        Req.Request.put_new_header(request, "if-modified-since", http_datetime_string)
```
wojtekmach, line 468:
```suggestion

        Req.Request.put_new_header(request, "if-modified-since", http_datetime_string)
```

wojtekmach:
Thank you! I've temporarily brought this back (33f54160c0ee13eabc5f2a65118e45de07c1bcec) with a deprecation warning. I plan to drop this for real in v0.5.

lau:
Great!

PR #180, opened 2023-04-23
[Proposal] Allow `:finch_request` to perform the underlying request

zachallaun:
Hi there! This is basically a proposal with code: Add a `:finch_exec_request` option to the default Finch adapter in order to enable custom plugins/steps to implement streaming response bodies.

[This recent ChatGPT streaming example](https://github.com/wojtekmach/mix_install_examples/pull/22) got me thinking about how the example could be implemented in Req. Finch provides support for streaming responses with `Finch.stream/5`, but the `run_finch` adapter always calls `Finch.request/3`. One could create their own `run_finch_stream` adapter, but they would be duplicating quite a lot of effort. The idea of this proposal is that an optional `:finch_exec_request` function would be provided with the Finch request, name, and options, and would return either a Req response or an exception.

Using this branch, it's fairly straightforward to write a plugin that allows for streaming responses. [Here's a Livebook notebook](https://gist.github.com/zachallaun/88aed2a0cef0aed6d68dcc7c12531649) showing how it might be done. There are some issues with resource handling (no back pressure, no explicit mechanism to cancel the request), but I believe this still opens up some useful options for plugins.

### DIFF
--- a/lib/req/response.ex
+++ b/lib/req/response.ex
@@ -30,8 +30,13 @@ defmodule Req.Response do
   @doc """
   Returns a new response.
   """
-  @spec new(options :: keyword()) :: t()
-  def new(options \\ []) do
+  @spec new(options :: keyword() | map() | struct()) :: t()
+  def new(options \\ [])
+
+  def new(options) when is_list(options), do: new(Map.new(options))
+
+  def new(options) do
+    options = Map.take(options, [:status, :headers, :body])
     struct!(__MODULE__, options)
   end
 
--- a/lib/req/response.ex
+++ b/lib/req/response.ex
@@ -29,6 +29,18 @@ defmodule Req.Response do
 
   @doc """
   Returns a new response.
+
+  Expects a keyword list, map, or struct containing the response keys.
+
+  ## Example
+
+      iex> Req.Response.new(status: 200, body: "body")
+      %Req.Response{status: 200, headers: [], body: "body"}
+
+      iex> finch_response = %Finch.Response{status: 200}
+      iex> Req.Response.new(finch_response)
+      %Req.Response{status: 200, headers: [], body: ""}
+
   """
   @spec new(options :: keyword() | map() | struct()) :: t()
   def new(options \\ [])
--- a/test/req/steps_test.exs
+++ b/test/req/steps_test.exs
@@ -1201,7 +1201,7 @@ defmodule Req.StepsTest do
     assert Req.request!(plug: plug, json: %{a: 1}).body == "ok"
   end
 
-  test "run_finch: :finch_request", c do
+  test "run_finch: :finch_request with arity 1 (deprecated)", c do
     Bypass.expect(c.bypass, "GET", "/ok", fn conn ->
       Plug.Conn.send_resp(conn, 200, "ok")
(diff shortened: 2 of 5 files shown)


wojtekmach:
This is a nice idea but I need to think about it more. I'm also not sure if we should have both :finch_request and :finch_exec_request. In other words, the latter can be used in place of the former with only slightly more code. Say, we had `finch_request: fun` and `fun` is:

```diff
- fun = & &1
+ fun = fn finch_request, finch_name, finch_options ->
+   case Finch.request(finch_request, finch_name, finch_options) do
+     {:ok, response} -> Req.Response.new(response)
+     {:error, exception} -> exception
+   end
+ end
```

That's not too bad?

If that sounds reasonable, some notes:

1. We'd deprecate `finch_request: fn request ->` in favour of this.
2. Perhaps contract should be even more flexible: `fn request, finch_request, finch_name, finch_options -> {request, response | exception}` but not sure what'd be the use case.
3. `Req.Response.new` will need to start accepting maps and structs which I'm fine with.

@aselder you needed :finch_request in the past. Any thoughts about this?

zachallaun:
Agreed on all the above. I think you’re already thinking this, but we can maintain backwards compatibility for some time by using the old `:finch_request` behaviour if the function is of arity 1 and the new behaviour if it’s 3/4. (I thought about also passing the req request but decided not to because I couldn’t really think of a use-case, since that can be modified in non-adapter steps.)

zachallaun:
[Here's what it might look like](https://github.com/wojtekmach/req/compare/main...zachallaun:req:finch-exec-request-2) if we just go with `:finch_request`. If this direction is preferable, I'll push to this branch.

lib/req/response.ex
```diff
@@ -29,9 +29,26 @@ defmodule Req.Response do
...
   """
-  @spec new(options :: keyword()) :: t()
-  def new(options \\ []) do
+  @spec new(options :: keyword() | map() | struct()) :: t()
+  def new(options \\ [])
+
+  def new(options) when is_list(options), do: new(Map.new(options))
+
+  def new(options) do
+    options = Map.take(options, [:status, :headers, :body])
```
zachallaun, line 51: I figured that this should not take `:private` to force using the `get/put_private` interface.

zachallaun:
Pretty sure the CI failures are unrelated to this PR. I'm able to replicate the failures running the workflows locally both on this branch and on `main`.

test/req/steps_test.exs
```diff
@@ -1213,10 +1213,50 @@ defmodule Req.StepsTest do
...
+    fun = fn req, _finch_request, _finch_name, _finch_opts ->
+      {req, %ArgumentError{message: "exec error"}}
+    end
+
+    assert_raise ArgumentError, "exec error", fn ->
+      Req.get!(c.url, finch_request: fun, retry: :never)
+    end
+  end
+
+  test "run-finch: :finch_request with invalid return", c do
```
wojtekmach, line 1252:
```suggestion
  test "run_finch: :finch_request with invalid return", c do
```

(thanks for this test btw!)

wojtekmach:
Thank you!

PR #183, opened 2023-05-09
Fix autolink reference for `Req.new/1` in README

Auralcat:
Currently the link to `Req.new/1` in the README points to`https://hexdocs.pm/req/Req.html#get!/2`:

demo-bug

This PR is to fix the link reference so people can go to the correct page in HexDocs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 [`Req.request/1`]: https://hexdocs.pm/req/Req.html#request/1
-[`Req.new/1`]:    https://hexdocs.pm/req/Req.html#get!/2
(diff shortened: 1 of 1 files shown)


wojtekmach:
Oh, good find! Could you fix this line instead? https://github.com/wojtekmach/req/blob/1e2640734bd442a2f6438c41d9f05600fcfad519/README.md?plain=1#L178

Auralcat:
@wojtekmach Done in https://github.com/wojtekmach/req/commit/6b6f7b3e44bd2ef58786a9c2a4ef7f426e33b594

Thank you for pointing line 178, this makes things so much simpler, I can stop digging into GFM's code now.  :smile:

wojtekmach:
Thank you!

PR #184, opened 2023-05-13
fix 307/308 redirect behavior

### LINKED ISSUE #175: Bug: HEAD requests with a 307/308 response are followed up with a GET request
### The bug
According to [the docs](https://hexdocs.pm/req/Req.Steps.html#follow_redirects/1) and RFC spec for http, requests which receive a 307 or 308 response should follow the redirect while maintaining the same request method. For example, a HEAD request to `/testing-redirect` with a 307 response to `/redirected-here` should make the request `HEAD /redirected-here`. Instead, Req will change the verb to `GET`. 

### Reproduction Steps
1) Copy this test into `req_test.exs`

```
test "HEAD redirect", c do
    path1 = "/testing-redirect"
    path2 = "/redirected-here"

    Bypass.expect(c.bypass, "HEAD", path1, fn conn ->
      conn
        |> Plug.Conn.put_resp_header("location", path2)
        |> Plug.Conn.resp(307, "")
    end)

    Bypass.expect_once(c.bypass, "HEAD", path2, fn conn ->
      conn
      |> Plug.Conn.put_resp_header("foo", "bar")
      |> Plug.Conn.resp(200, "")
    end)

    Req.head!(c.url <> path1)
  end
```

2) Run `mix test` and the test will fail with the following response
```
mix test
Excluding tags: [:integration]

.......
11:07:59.095 [debug] follow_redirects: redirecting to /redirected-here

11:07:59.098 [error] retry: got response with status 500, will retry in 1000ms, 3 attempts left

11:07:59.098 [error] #PID<0.398.0> running Bypass.Plug (connection #PID<0.396.0>, stream id 2) terminated
Server: localhost:60288 (http)
Request: GET /redirected-here
** (exit) an exception was raised:
    ** (RuntimeError) route error
        (bypass 2.1.0) lib/bypass/plug.ex:28: Bypass.Plug.call/2
        (plug_cowboy 2.6.0) lib/plug/cowboy/handler.ex:11: Plug.Cowboy.Handler.init/2
        (cowboy 2.9.0) /Users/conor/src/github.com/wojtekmach/req/deps/cowboy/src/cowboy_handler.erl:37: :cowboy_handler.execute/2
        (cowboy 2.9.0) /Users/conor/src/github.com/wojtekmach/req/deps/cowboy/src/cowboy_stream_h.erl:306: :cowboy_stream_h.execute/3
        (cowboy 2.9.0) /Users/conor/src/github.com/wojtekmach/req/deps/cowboy/src/cowboy_stream_h.erl:295: :cowboy_stream_h.request_process/3
        (stdlib 3.17.2) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
```

Note the `Request: GET /redirected-here` which should instead be `Request: HEAD /redirected-here

### Possible fix

I think the issue here is that the method definition which pattern matches on the `request.status` property fails because we need match on `response.status` instead
https://github.com/wojtekmach/req/blob/36fa23632e1e77299aea5cdd8d76e40203f1e982/lib/req/steps.ex#L1166-L1168

Should instead be something like
```
defp put_redirect_request_method(request, response) when response.status in 307..308, do: request

defp put_redirect_request_method(request, response), do: %{request | method: :get}
```

Happy to contribute a PR to fix this if it sounds correct!

srowley:
Fixes #175. (my bad)

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -1217,7 +1217,7 @@ defmodule Req.Steps do
     request
     |> remove_params()
     |> remove_credentials_if_untrusted(location_trusted, location_url)
-    |> put_redirect_request_method()
+    |> put_redirect_request_method(response.status)
     |> put_redirect_location(location_url)
   end
 
@@ -1234,9 +1234,9 @@ defmodule Req.Steps do
     put_in(request.url, location_url)
   end
 
-  defp put_redirect_request_method(request) when request.status in 307..308, do: request
+  defp put_redirect_request_method(request, status) when status in 307..308, do: request
 
-  defp put_redirect_request_method(request), do: %{request | method: :get}
+  defp put_redirect_request_method(request, _), do: %{request | method: :get}
 
   defp remove_credentials_if_untrusted(request, true, _), do: request
 
--- a/test/req/steps_test.exs
+++ b/test/req/steps_test.exs
@@ -617,6 +617,20 @@ defmodule Req.StepsTest do
            end) =~ "[debug] follow_redirects: redirecting to #{c.url}/ok"
   end
 
+  test "follow_redirects: 307..308", c do
+    Bypass.expect(c.bypass, "POST", "/redirect", fn conn ->
+      redirect(conn, 307, c.url <> "/ok")
+    end)
+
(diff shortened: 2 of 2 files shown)


wojtekmach:
Thank you!

PR #185, opened 2023-05-24
URL Encode the Location header when redirecting

robbyronk:
Mint will fail with invalid_request_target without this when encountering URLs in Location headers with spaces.


robbyronk:
Seems like upgrading mint (or something else) ended up fixing this! Cheers

PR #192, opened 2023-06-19
Set content-length when decompressing

tanguilp:
When decompressing, the `content-length` is left unchanged. As a consequence, it's no longer matches the body's length, and could confuse other steps or users.

This PR automatically sets `content-length` after decompression.

It could be discussed if it's worth adding it if it wasn't there in the first place. I'd argue it cannot hurt, and proxies routinely do this. Not only it doesn't hurt, but it can help a client know when the response has been fully recieved (should `Req` be used in a reverse-proxy one day) especially when using multiplexing (`Connection: keep-alive`).

Not sure if we should update the documentation to notice users about these changes.

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -861,6 +861,9 @@ defmodule Req.Steps do
   | zstd          | `:ezstd.decompress/1` (if [ezstd] is installed) |
   | _other_       | Returns data as is                              |
 
+  This step updates the following headers to reflect the changes:
+  - `content-legnth` is set to the length of the decompressed body
+
   ## Options
 
     * `:raw` - if set to `true`, disables response body decompression. Defaults to `false`.
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -862,7 +862,7 @@ defmodule Req.Steps do
   | _other_       | Returns data as is                              |
 
   This step updates the following headers to reflect the changes:
-  - `content-legnth` is set to the length of the decompressed body
+  - `content-length` is set to the length of the decompressed body
 
   ## Options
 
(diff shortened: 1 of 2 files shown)


tanguilp:
Asked about Tesla here: https://github.com/elixir-tesla/tesla/issues/598

wojtekmach (approved):

wojtekmach:
Thank you! Could you rebase on main and mention this behaviour in docs?

lib/req/steps.ex
```diff
@@ -817,6 +817,9 @@ defmodule Req.Steps do
   | zstd          | `:ezstd.decompress/1` (if [ezstd] is installed) |
   | identity      | Returns data as is                              |
 
+  This step updates the following headers to reflect the changes:
+  - `content-legnth` is set to the length of the decompressed body
+
```
tanguilp, line 822: We'll have to add `content-encoding` is removed, but in another PR. Hence the list with one item for now

lib/req/steps.ex
```diff
@@ -818,7 +818,7 @@ defmodule Req.Steps do
   | identity      | Returns data as is                              |
 
   This step updates the following headers to reflect the changes:
-  - `content-legnth` is set to the length of the decompressed body
+  - `content-length` is set to the length of the decompressed body
```
tanguilp: We'll have to add that `content-encoding` is removed, but this will be done in [another PR](https://github.com/wojtekmach/req/pull/190#pullrequestreview-1494951901). Hence the list with one item for now

tanguilp:
Reopening after fixing conflicts (it was out-of-sync).

In the current state, this PR is a prerequisite for https://github.com/wojtekmach/req/pull/190 but let me know if you wanna merge in a different order (or merge at all).

yordis:
@tanguilp @wojtekmach do you mind code reviewing https://github.com/elixir-tesla/tesla/pull/606 🙏🏻 🙏🏻 🙏🏻 🙏🏻 🙏🏻 🙏🏻 🙏🏻

Issue #194, opened 2023-06-24
Support `response[header]`


wojtekmach:
This should only accept string keys, return header value when there was exactly one such header, nil if there were none, and raise otherwise pointing to using Req.Response.get_header/2 instead.

PR #199, opened 2023-07-11
Raise when trying to fetch response header with multiple values

andreyuhai:
Hi! 👋

I came across this library on Twitter, started using it and wanted to see if I could start contributing somehow.
This seemed like the only missing part for #194 so I wanted to create a PR, unless I misunderstood something. :)


wojtekmach:
Thank you!

wojtekmach:
We still need to document the feature but let's hold off because I have heard from some people that it might not be such a good idea after all. No rush!

PR #203, opened 2023-07-16
add :retry_after_filter option

DaTrader:


wojtekmach:
Thank you for the PR.

I thought we'd just change:

https://github.com/wojtekmach/req/blob/v0.3.10/lib/req/steps.ex#L1375

to handle both 429 and 503. Why is that not enough?

wojtekmach:
if it's important to have total control over using or not using retry-after header, I'd be ok with passing {request, response} to the retry_delay function so someone could do:

```elixir
retry_delay: fn {request, response} ->
  case get_header(response, "retry-after") do
    [number_or_date] ->
      case Integer.parse(number_or_date) do
        {seconds, ""} ->
          seconds ** 1000

        :error ->
          # parse_http_date(...)
      end
    [] ->
      (2 ** request.private.req_retry_count) * 1000
  end
end
```

would that help?

DaTrader:
1. It's not that 429 and 503 would not be enough, it's that in this particular case I, for instance, only need it for the 503 and not for the 429. Maybe someone else will need it for both, maybe with different delay logic etc. That's why we need some flexibility and why hard coded status codes may not be such a good idea.

2. As for the suggested function, yes, generally it would help. Although this above example as such requires the user to replicate some of the library code. Personally, I'd go for more code reuse. The `get_header/2` is not expensive so the `retry-after` can be supplied to the function as well, resulting in something like this (I'm not into implementation now, just the interface):

```elixir
  @type falsey() :: false | nil
  @type retry_delay_result() :: integer() | falsey()

  @type retry_delay_fun() ::
          ( retry_count :: integer() -> retry_delay_result())
          | ( Req.Request.t(), Req.Response.t(), retry_after :: integer() -> retry_delay_result())

  # usage in my case:
  Req.get!( url, fn _, response, retry_after ->
    response.status == 503 && retry_after
  end)
```

or better yet:

```elixir
  @type retry_delay_args() ::
          %{
            required( :request) => Req.Request.t(),
            required( :response) => Req.Response.t(),
            required( :retry_after) => integer()
          }

  @type retry_delay_fun() ::
          ( retry_count :: retry_count() -> retry_delay_result())
          | ( retry_delay_args() -> retry_delay_result())

  # usage, again in my case:
  Req.get!( url, & &1.response.status == 503 && &1.retry_after)
```

DaTrader:
@wojtekmach Please check if the suggested function interface is ok with you and let me know if you want to implement it yourself or I can make a PR.

wojtekmach:
Thanks for the update, I need some more time to think about it.

DaTrader:
@wojtekmach Hey, any chance to confirm this or make an alternative suggestion for I kind of need this feature quite soon.

Thanks

wojtekmach:
In your proposal what happens if the retry_delay fun returns false?

The problem with `%{request:, response:, retry_delay:}` as input is the retry might have been caused by getting an exception. We could extend the map to have an :exception key too. And then either :response OR :exception field would be non-nil but I'm not sure if such API would be easy to use.

My proposal is to add a `fun(req, resp_or_exception, delay) -> milliseconds` where `delay` can be nil (ie we got an exception or a response without the header)

A nice thing about a 3-arity function is that it's backwards compatible. We can deprecate passing 1-arity function in favour of the new one.

WDYT? If that sounds good a PR is welcome.

wojtekmach:
Alternatively it's fun(req, resp) but we add a `Req.Response.get_retry_after(unit \\ :millisecond) :: non_neg_integer() | nil` and the users can grab the value of the header themselves.

According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After the header might also appear on HTTP 301 Moved Permanently (which we currently do not take into account) so this function can be more generally useful in users own code too.

I think I slightly prefer this over sending `fun(req,resp,delay)` where the latter can be nil. WDYT?

DaTrader:
> In your proposal what happens if the retry_delay fun returns false?

falsey -> Retrial stops immediately.

> The problem with `%{request:, response:, retry_delay:}` as input is the retry might have been caused by getting an exception. We could extend the map to have an :exception key too. And then either :response OR :exception field would be non-nil but I'm not sure if such API would be easy to use.

My suggestion is based on the same presumptions as the `Steps.retry/1` `:retry` fun arity 1 option. Therefore, yes, we can have it accept a response or an exception.

I agree with not using a map as a single argument. It was probably just me going a bit too far to make it as convenient for the user as possible (which in the end turns out not be so).

> Alternatively it's fun(req, resp) but we add a `Req.Response.get_retry_after(resp, unit \\ :millisecond) :: non_neg_integer() | nil` and the users can grab the value of the header themselves.

That one would be a plain code reuse approach, but I personally like the other approach better as this one is making the user implement the loop the library could easily provide for them (as it does in other cases).

> According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After the header might also appear on HTTP 301

I don't think we should worry bout the status codes if we provide the response to the use function (and we do).

> I think I slightly prefer this over sending `fun(req,resp_or_error,delay)` where the 3rd argument will sometimes be nil for responses and _always_ nil for errors. WDYT?

We can easily have the function receive `:error` as the third argument if an exception is provided. There's nothing wrong about it since what we're really talking about here is coping with the anonymous function limitation of having a fixed number of arguments.

The altered proposal would then be as follows:

```elixir
@type retry_delay_fun() :: 
  ( retry_count :: retry_count() -> retry_delay_result()) # for backward compatibility
  | ( Req.Request.t(), Req.Response.t() | Exception.t(), retry_after :: integer() | :error -> retry_delay_result())
```

wojtekmach:
> falsey -> Retrial stops immediately.

About falsey, retry _delay_ option controlling whether retries _happen_ is to me mixing concerns. If anything perhaps I'd do the opposite, :retry option could return true/false or delay. But not sure how flexible that is.

> retry_after :: integer() | :error

this needs to also have `| nil` for when the response does not have retry-after header. But then I think it's a bit much, I think I'd be ok with just `non_neg_integer() | nil` then, nil can happen when we got the responds without the header or the exception.

yordis:
In general, I would say avoid booleans. Could you return `:stop` or `{:continue, 10}` or whatever union of things that are not true|false (boolean)?

Something that defines an intent.

DaTrader:
> retry _delay_ option controlling whether retry _happen_ is to me mixing concerns. If anything perhaps I'd do the opposite, :retry option could return true/false or delay. But not sure how flexible that is.

I don't see it that way because here we come down to `:retry` and `:retry_delay` options being mutually exclusive i.e. two distinct alternatives each of which then must be complete and have its own way of saying when to stop. Your suggestion confirms that, it's just that you suggest moving it into the `:retry` option which I wouldn't have anything against if the flexibility as kept but then we risk the backward compatibility.

> this needs to also have `| nil` for when the response does not have retry-after header. But then I think it's a bit much, I think I'd be ok with just `non_neg_integer() | nil` then, nil can happen when we got the response without the header or the exception.

I now see what you meant by adding the `Req.Response.get_retry_after/2`. Ok, we can do it this way so the user function only receives the request and response or exception.

wojtekmach:
:retry is a `fn req`, we can always add a `fn req, resp_or_error` in a backwards compatible way. Again not sure this is the right path.

Actually going back to what you wrote before:

```elixir
Req.get!(url, retry_delay: & &1.response.status == 503 && &1.retry_after)
```

is this _really_ what you want? Why wouldn't you retry on 429 (too many requests), 504 (gateway timeout) or exceptions like `%Mint.TransportError{reason: :timeout}`?

DaTrader:
> Why wouldn't you retry on 429 (too many requests), 504 (gateway timeout) or exceptions like `%Mint.TransportError{reason: :timeout}`?

Because I have to abide by the API specs that explicitly state to retry just on 503. It's like you have much choice if you're building just the front end.

DaTrader:
.. and this is why I am insisting so much on the status code interpretation flexibility.

wojtekmach:
Got it, thank you.

On one hand keeping :retry and :retry_delay orthogonal keeps things simpler, more predictable, and there are no weird interactions between the two. On the other hand if you'd need something really custom then it's not ideal to have some logic in :retry and them some more logic in :retry_delay and most likely duplicating some conditionals between the two.

So the only way I can see it is this:

- `retry: fn req` is deprecated in favour of `retry: fn req, resp_or_error`
- If `retry: fn req, resp_or_error -> false | nil` we don’t retry
- If `retry: fn req, resp_or_error -> true`, we retry with a delay that from response with “retry-after” header, otherwise we use `:retry_delay` (defaults to simple exp backoff)
- If `retry: fn req, resp_or_error -> {:delay, milliseconds}` we delay for that time. If :retry_delay is set, we raise because in this particular case these are mutually exclusive.

And so for your use case we'd have:

```elixir
retry: fn
  _req, %Req.Response{status: 503} = resp ->
    if delay = Req.Response.get_retry_after(resp) do
      {:delay, delay}
    end

  _req, _resp_or_error ->
    false
end
```

wdyt?

wojtekmach:
I agree in principle specific atoms are better than booleans but I can't think of a good replacement for `true` nor particularly like `:halt` or `:stop` for this use case.

Making this extremely convenient to use is not the highest priority for me but the falsy return you were pushing for is obviously especially useful with `if` like shown above.

Btw, I just realised that for your use case this is the shortest (again LOC not being the highest priority):

```elixir
retry: fn _req, resp_or_error -> match?(%Req.Response{status: 503}, resp_or_error end
```

DaTrader:
> That is unless the frontend is like you can retry only if it's a 503 AND we send a Retry-After sigh.

It is, but never mind. This is now ok (the former usage example)

DaTrader:
So, we're settled then? I can go with a new PR in the following days for these specs?

wojtekmach:
Yup please send the PR. Thanks!

DaTrader:
> * If `retry: fn req, resp_or_error -> true`, we retry with a delay from response with “retry-after” header, otherwise we use `:retry_delay` (defaults to simple exp backoff)

Note that this differs from your current behavior. Currently, if a `:retry` `fun/1` returns true, the `retry-after` header will only be considered if the status code is 429. My recommendation is to keep it this the same or it may become a breaking change for some users.

wojtekmach:
Yes if users blindly change behaviour their code might break but the whole point of introducing a new function is to have an opportunity to do things right. We can help users by explaining things in the documentation and in the changelog. Please handle retry-after on 503s as well. Unless I'm missing some downsides of doing so?

wojtekmach:
Closing in favour of #226.

PR #206, opened 2023-07-24
fixes #205 : incorrect "content-range" headers instead of "range"

### LINKED ISSUE #205: Content Range - put_range step has wrong implementation
https://github.com/wojtekmach/req/blob/v0.3.11/lib/req/steps.ex#L416

the docs say that `put_range/1` sets the `"content-range"` header. It actually, sets the `"range"` header as seen from code above. 

Either the docs need to change or the `put_range/1`. 

Willing to raise a PR.

TwistingTwists:
As discussed in #205 , corrected "content-range" headers.

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -413,11 +413,11 @@ defmodule Req.Steps do
   """
   @doc step: :request
   def put_range(%{options: %{range: range}} = request) when is_binary(range) do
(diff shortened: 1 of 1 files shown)


wojtekmach:
See https://github.com/wojtekmach/req/issues/205#issuecomment-1648255726

PR #208, opened 2023-07-28
Fix typespecs for some functions

RobinBoers:
Hi, I was using this library and dialyzer started to complain:

```dialyzer
lib/dummy/dummy.ex:94:call
The function call will not succeed.

Req.put([
  {:body, binary()} | {:headers, [{<<_::72>>, _}, ...]} | {:url, <<_::8, _::size(1)>>},
  ...
])

will never return since the 1st arguments differ
from the success typing arguments:

(
  binary()
  | %{
      :__struct__ => Req.Request | URI,
      :adapter => (... -> any),
      :authority => URI.authority(),
      :body =>
        nil
        | binary()
        | maybe_improper_list(
            binary() | maybe_improper_list(any(), binary() | []) | byte(),
            binary() | []
          ),
      :current_request_steps => _,
...
```

So I took a look at the source code, and saw that for some functions the `keyword()` type was missing in the typespec, while all those functions had a clause to handle keyword lists, and the docs clearly said a keyword list was accepted.

So I added them, and also made the docs for `get/2` consistent with the rest.

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -383,6 +383,12 @@ defmodule Req do
   @doc """
   Makes a GET request.
 
+  `request` can be one of:
+
+    * a `String` or `URI`;
+    * a `Keyword` options;
+    * a `Req.Request` struct
+
   See `new/1` for a list of available options.
 
   ## Examples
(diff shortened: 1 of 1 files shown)


wojtekmach:
Thank you!

PR #224, opened 2023-08-20
`Req.Request`, `Req.Response`: Change headers to be maps

### LINKED ISSUE #223: Change headers to maps
Currently Req request and response headers are lists of tuples, e.g.: `[{"foo", "bar"}]`. This is a standard across the ecosystem (with minor difference that some Erlang libraries use charlists instead of binaries.) This particular representation makes sense because the same header might appear multiple times, see [RFC 9110 § 5.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.2).

There are some problems with this particular choice though:

1. We cannot use `headers[key]`
2. We cannot use pattern matching

In short, this representation isn't very ergonomic to use.

## 1. Headers as Maps

Both of these problems can be solved by switching the representation to maps of string keys and list values:

```diff
- [{"foo", "bar"}]
+ %{"foo" => ["bar"]}
```

This is similar to [Go's net/http `Header` type](https://pkg.go.dev/net/http#Header).

Let's see some example rewrites:

```diff
- [value] = Req.Response.get_header(response, "content-type")
+ [value] = response.headers["content-type"]
```

or in certain situations even better:

```elixir
case response do
  %{headers: %{"content-type" => ["application/json" <> _]}} ->
end
```

## Option 2: Headers as Structs

An interesting idea is to go further and have a struct:

```elixir
defmodule Req.Headers do
  defstruct [:map]
end

resp = Req.get!("https://elixir-lang.org")
resp.headers["content-type"] #=> ["text/html"]
resp.headers["Content-Type"] #=> ["text/html"] (RFC says they need to be treated case-insensitively!)
```

and we _still_ could pattern match:

```elixir
case response do
  %{headers: %{map: %{"content-type" => "text/html" <> _}}} ->
end
```

however here we cannot treat the names case insensitively so it's probably better to continue assuming they are lowercase (and enforce it: #196).

The struct gives us opportunity for slightly different Access implementation however:

```diff
- resp.headers["content-type"] #=> ["text/html"]
+ resp.headers["content-type"] #=> "text/html"
```

and the semantics would be similar to [Go's `Header.Get()`](https://pkg.go.dev/net/http#Header.Get):

> Get gets the first value associated with the given key. 

and then we'd have something like [`Header.Values()`](https://pkg.go.dev/net/http#Header.Values).

With structs we can implement protocols. For example, we may want to have an enumerable implementation for interop:

```elixir
Enum.to_list(resp.headers) #=> [{"content-type", "text/html"}, ...]
```

And perhaps Collectable too:

```elixir
Enum.into([{"content-type", "text/html"}], resp.headers)
```

Basically, we have extensibility.

All that being said, while I thing these are nice things, I feel like structs are a more complicated solution than plain maps and I'm not sure they give us significant benefits at the moment.

wojtekmach:
Closes #223

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,44 @@
 
 ## HEAD
 
+  * Change `request.headers` and `response.headers` to be maps.
+
+    Previously headers were lists of name/value tuples, e.g.:
+
+        [{"content-type", "text/html"}]
+
+    Now they are maps of string names and lists of values, e.g.:
+
+        %{"content-type" => ["text/html"]}
+
+    This is a major breaking change. If you cannot easily update your app
+    or your dependencies, do:
+
+        # config/config.exs
+        config :req, legacy_headers_as_lists: true
+
+    This legacy fallback will be removed on Req 1.0.
+
+  * Make `request.registered_options` internal representation private.
+
+  * Make `request.options` internal representation private.
+
+    Currently `request.options` field is a map but it may change in the future.
+    One possible future change is using keywords lists internally which would
+    allow, for example, `Req.new(params: [a: 1]) |> Req.update(params: [b: 2])`
+    to keep duplicate `:params` in `request.options` which would then allow to
+    decide the duplicate key semantics on a per-step basis. And so, for example,
+    [`put_params`] would _merge_ params but most steps would simply use the
+    first value.
+
+    To have some room for manoeuvre in the future we should stop pattern
+    matching on `request.options`. Calling `request.options[key]`,
+    `put_in(request.options[key], value)`, and
+    `update_in(request.options[key], fun)` _is_ allowed.
+    New functions [`Req.Request.get_option/3`],
+    [`Req.Request.fetch_option!/2`], and [`Req.Request.delete_option/1`] have been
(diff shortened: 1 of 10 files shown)


florish:
Hi @wojtekmach, thanks for all the work put into `Req`!

We're in the processing of migrating from 0.3 to 0.4, and in the process I assumed that the `:headers` option in `Req.new/1` and `Req.update/2` would also have been changed to expect map data instead of lists. E.g.:

```
req = Req.new(url: "https://example.com", headers: %{"accept" => ["application/json"]})
Req.update(req, headers: %{"foo" => ["bar"]})
```

Is there a specific reason that in both cases, the `:headers` values are still lists in 0.4.x?

wojtekmach:
Both are supported! (If maps are not, it’s a bug.) the idea is passing eg `accept: “application/json”` is much more concise than the equivalent map with, notably, list of values.

wojtekmach:
In hindsight this is not necessarily super obvious so a PR that updates docs is welcome as always!

florish:
@wojtekmach See #241 for PR, feel free to change anything you like!

PR #241, opened 2023-09-04
Use maps for request header examples

florish:
As discussed here: 

https://github.com/wojtekmach/req/pull/224#issuecomment-1705159795

As a couple of `iex` tests learned my that `headers: %{"key" => "value"}` maps are also accepted and merged correctly (i.e. no need to explicitly wrap `"value"` in `["value"`]), using maps in the examples felt like an improvement, as they are closer to the `%Req.Request{}` typespec.

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -113,7 +113,7 @@ defmodule Req do
 
     * `:url` - the request URL.
 
-    * `:headers` - the request headers as a key-value map or key-value tuple list. 
+    * `:headers` - the request headers as a `{key, value}` enumerable (e.g. map, keyword list). 
 
       The header names should be downcased.
 
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -113,7 +113,9 @@ defmodule Req do
 
     * `:url` - the request URL.
 
-    * `:headers` - the request headers. The header names should be downcased.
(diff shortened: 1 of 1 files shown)


lib/req.ex
```diff
@@ -113,7 +113,9 @@ defmodule Req do
 
     * `:url` - the request URL.
 
-    * `:headers` - the request headers. The header names should be downcased.
+    * `:headers` - the request headers as a key-value map or key-value tuple list.
```
wojtekmach, line 116: strictly speaking they can be any _enumerable_ of `{name, value_or_values}` pairs

i.e. this all works:

```elixir
foo: "bar"

foo: ["bar"]

%{foo: "bar"}

%{foo: ["bar"]}
```

I'm not sure how to best describe it.

florish: Ah, I'm not the only one having trouble with that, then 😄 

```suggestion
    * `:headers` - the request headers as a `{key, value}` enumerable (e.g. map, keyword list). 
```

Would this be better?

florish: Additional note: given that in request headers it rarely makes sense to pass the same header multiple times, my thought was that it would be OK to omit explicit examples with list-wrapped header values.

Please correct me on this if needed!

wojtekmach:
Thank you, sorry for the delay!

PR #252, opened 2023-09-26
Add `:finch_private` option to put into Finch.Request

### LINKED ISSUE #251: Finch: Add option for storing private options onto Finch request (for telemetry)
I'm looking to add some private data to Finch requests so I can catch them in [telemetry events](https://hexdocs.pm/finch/Finch.Telemetry.html).

Currently, the finch options are limited to `[:receive_timeout, :pool_timeout]` and it looks like I'd need to rebuild much of Req's finch adapter (especially re: streaming) to be able to inject private options onto the `%Finch.Request{}` via the `:finch_request` option myself. 

For non-streaming requests, it looks simple enough to add a `:finch_request` option and not miss much of Req's features:

https://github.com/wojtekmach/req/blob/59db3e26e3baab3b1af95c27bcc5045e5b3b0aff/lib/req/steps.ex#L699-L701

https://github.com/wojtekmach/req/blob/59db3e26e3baab3b1af95c27bcc5045e5b3b0aff/lib/req/steps.ex#L834-L839

This is reasonable for me bring into the app for normal requests, for example:

```elixir
# My Req shim provides setup with default options, and this would be one of them:
defp put_finch_request(opts) do
  finch_private = Keyword.get(opts, :finch_private, %{})

  Keyword.put_new(:finch_request, fn _req, finch_req, finch_name, finch_options ->
    finch_req = Enum.reduce(finch_private, finch_req, fn {k, v}, finch_req_acc ->
      Finch.Request.put_private(finch_req_acc, k, v)
    end)

    case Finch.request(finch_req, finch_name, finch_options) do
      {:ok, response} -> {req, Req.Response.new(response)}
      {:error, exception} -> {req, exception}
    end
  end)
end
```

but this becomes much more tedious with streaming requests which I don't want to rebuild or miss out on if a fellow developer decides to a stream a response (but sacrifice our telemetry), as there is a lot more that Req is doing to handle various scenarios.

The other 3 clauses branch into ~ 120 LoC that feels less reasonable to bring app-side:

https://github.com/wojtekmach/req/blob/59db3e26e3baab3b1af95c27bcc5045e5b3b0aff/lib/req/steps.ex#L699-L711

---

I'm hoping maybe we can allow an option to insert Finch private metadata, around here:
 
https://github.com/wojtekmach/req/blob/59db3e26e3baab3b1af95c27bcc5045e5b3b0aff/lib/req/steps.ex#L679-L682

I'd love to contribute a `finch_private: [] | %{}` option to the Req Finch adapter which will reduce them onto the finch request via `Finch.Request.put_private/3`, maybe around here: https://github.com/wojtekmach/req/blob/59db3e26e3baab3b1af95c27bcc5045e5b3b0aff/lib/req/steps.ex#L675-L677

This Finch request will be carried through all of the response types and be easier for Req users to catch telemetry events with custom metadata.

dbernheisel:
Resolve #251 

This adds the `:finch_private` option which will allow adding private metadata to the Finch request when built so it travels through Req's finch adapter. This is helpful for putting custom metadata in for handling in `:telemetry` event handlers.

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -293,6 +293,9 @@ defmodule Req do
 
     * `:unix_socket` - if set, connect through the given UNIX domain socket.
 
+    * `:finch_private` - a map or keyword list of private metadata to add to the Finch request. May be useful
+      for adding custom data when handling telemetry with `Finch.Telemetry`.
+
     * `:finch_request` - a function that executes the Finch request, defaults to using `Finch.request/3`.
 
   ## Examples
@@ -350,6 +353,7 @@ defmodule Req do
           :plug,
           :finch,
           :finch_request,
+          :finch_private,
           :connect_options,
           :inet6,
           :receive_timeout,
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -587,6 +587,9 @@ defmodule Req.Steps do
 
     * `:unix_socket` - if set, connect through the given UNIX domain socket.
 
+    * `:finch_private` - a map or keyword list of private metadata to add to the Finch request. May be useful
+      for adding custom data when handling telemetry with `Finch.Telemetry`.
+
     * `:finch_request` - a function that executes the Finch request, defaults to using `Finch.request/3`.
 
       The function should accept 4 arguments:
@@ -675,6 +678,7 @@ defmodule Req.Steps do
     finch_request =
       Finch.build(request.method, request.url, request_headers, body)
       |> Map.replace!(:unix_socket, request.options[:unix_socket])
+      |> add_private_options(request.options[:finch_private])
 
     finch_options =
       request.options |> Map.take([:receive_timeout, :pool_timeout]) |> Enum.to_list()
@@ -836,6 +840,15 @@ defmodule Req.Steps do
       {:ok, response} -> Req.Response.new(response)
       {:error, exception} -> exception
     end
+  end
+
+  defp add_private_options(finch_request, nil), do: finch_request
+
+  defp add_private_options(finch_request, private_options)
+       when is_list(private_options) or is_map(private_options) do
+    Enum.reduce(private_options, finch_request, fn {k, v}, acc_finch_req ->
+      Finch.Request.put_private(acc_finch_req, k, v)
+    end)
   end
 
   defp finch_fields_to_map(fields) do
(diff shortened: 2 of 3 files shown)


aselder:
Looks good to me... Almost identical to what I was planning writing.

wojtekmach:
Thank you!

wojtekmach:
Sorry, I can't help but wonder if we can have a more general mechanism.

Maybe a Finch.build hook?

```elixir
Req.new(
  finch_build: fn method, url, headers, body, options ->
    Finch.build(method, url, headers, body, options)
    |> Finch.Request.put_private(:foo, :bar)
  end
)
```

A nice thing is we could could set _any_ future (though unlikely) Finch struct field, inspect what we have, etc. Or something like:

```elixir
Req.new(
  finch_after_build: &Finch.Request.put_private(&1, :foo, :bar)
)
```

If anything the former seems more idiomatic but maybe both are a bit odd. Just to be clear I'm fine with moving forward with `:finch_private` as it's very self-explanatory but just thinking out loud here. cc @zachallaun @sneako

wojtekmach:
To give some more context, we used to be able to do:

```elixir
finch_request: &Finch.Request.put_private(&1, :foo, :bar)
```

and we deprecated that in favour of:

```elixir
finch_request: fn req, finch_req, finch_name, finch_options ->
  case Finch.request(finch_req, finch_name, finch_options) do
    {:ok, reps} -> {req, Req.Response.new(resp)}
  end
end
```

dbernheisel:
WDYT of a generic option `request_options`? When provided, Req pops out or merges its conveniences (like `unix_socket`), and puts the remaining options into Finch's private space via `put_private`.

Not choosing `finch_request_options` because I figure other adapters could take advantage of it as well.

wojtekmach:
@dbernheisel I'm worried about conflating options with private, they are quite distinct. An option that could be used by any adapter is pretty interesting, however. I'll think about it some more.

@zachallaun yeah un-deprecating `finch_request: &fun/1` and deprecating `&fun/4` seems like the most sensible path. Finch.request/3 uses Finch.stream_while/5 under the hood and I think Req will only use the latter so we have a single point where we hit the network.

I haven't decided if I'll revert this because just passing `:finch_private` is quite ergonomic to use. If so, I'll leave a message in this thread.

Thanks all and sorry for back and forth.

PR #255, opened 2023-10-06
Remove current_request_steps in Req.Request

thenrio:
Just use request_steps.

### DIFF
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -372,7 +372,6 @@ defmodule Req.Request do
             error_steps: [],
             private: %{},
             registered_options: MapSet.new(),
-            current_request_steps: [],
             into: nil,
             async: nil
 
@@ -615,8 +614,7 @@ defmodule Req.Request do
   def append_request_steps(request, steps) do
     %{
       request
-      | request_steps: request.request_steps ++ steps,
-        current_request_steps: request.current_request_steps ++ Keyword.keys(steps)
+      | request_steps: request.request_steps ++ steps
     }
   end
 
@@ -633,8 +631,7 @@ defmodule Req.Request do
   def prepend_request_steps(request, steps) do
     %{
       request
-      | request_steps: steps ++ request.request_steps,
-        current_request_steps: Keyword.keys(steps) ++ request.current_request_steps
+      | request_steps: steps ++ request.request_steps
     }
   end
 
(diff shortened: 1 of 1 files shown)


wojtekmach:
Thank you for the PR. It seems we are missing tests because the build should have failed as we needed this design for things like not replying base_url+url on retries, etc. I’ll investigate. Maybe it is not needed after all!

wojtekmach:
Note to self: revisit Req.Request.prepare/1 while at it.

thenrio:
Hi @wojtekmach, I noticed that `Req.Request.run_request/1` "consumes" the `current_request_steps` : https://github.com/wojtekmach/req/blob/main/lib/req/request.ex#L979.
Then we don't run `base_url` or `params` twice on `retry`, such as in https://github.com/wojtekmach/req/blob/main/test/req/steps_test.exs#L1587.
In fact, `retry` does not run any of the input `request_steps`, as `current_request_steps` is empty...

This commit "consumes" the `request_steps`, so we don't run `base_url` or `params` twice on `retry`.
We achieve same behavior with a list removed from memory and a lookup removed (https://github.com/wojtekmach/req/blob/7616d399914eb28377d9fa7fdb0b0a0273e181cb/lib/req/request.ex#L975).

Cheers.

wojtekmach:
Sorry for delay, this is tricky because the whole reason we introduced :current_request_steps was to avoid this:

```elixir
Req.new(params: [a: 1])
|> Req.Request.prepend_response_steps(
  test: fn {req, resp} ->
    req = put_in(req.current_request_steps, Keyword.keys(req.request_steps))
    {req, resp}
  end
)
|> Req.get!(
  retry_delay: 10,
  plug: fn conn ->
    dbg(conn.query_string)
    # conn.query_string #=> "a=1"
    # conn.query_string #=> "a=1&a=1"
    # conn.query_string #=> "a=1&a=1&a=1"
    # conn.query_string #=> "a=1&a=1&a=1&a=1"
    Plug.Conn.send_resp(conn, 500, "")
  end
)
```

(https://github.com/wojtekmach/req/issues/140)

But yes, on the flip side it causes https://github.com/wojtekmach/req/issues/296.

I haven't made up my mind yet which way to go, apologies for the delay.

wojtekmach:
Another data point: https://github.com/reisub/http_cookie/pull/3/files#diff-afcb5ad934b330c8e7ff684f4861d7a95409c5440099af7c8e47919b210f60c0R65.

wojtekmach:
Another one: https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/664

wojtekmach:
Closing in favour of #550. Thank you for this and apologies for the delay.

PR #256, opened 2023-10-06
Add step execution Telemetry

ethangunderson:
When debugging a latency increase in our Req clients we found it extremely useful to have the ability to create otel spans around Req step execution. We gain a similar level of granularity by attaching to Finch telemetry, but this helps us visualize everything that is happening in a request not just the connection pool and network bits.

What do you think?

### DIFF
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -476,7 +476,7 @@ defmodule Req.Request do
       iex> Req.Request.get_option_lazy(req, :b, fun)
       42
   """
-  @spec get_option_lazy(t(), atom(), (-> term())) :: term()
+  @spec get_option_lazy(t(), atom(), (() -> term())) :: term()
   def get_option_lazy(request, key, fun) when is_function(fun, 0) do
     Map.get_lazy(request.options, key, fun)
   end
@@ -707,7 +707,7 @@ defmodule Req.Request do
 
   @doc false
   def prepare(%{request_steps: [step | steps]} = request) do
-    case run_step(step, request) do
+    case run_step(step, request, :request) do
       %Req.Request{} = request ->
         request = %{request | request_steps: steps}
         prepare(request)
@@ -969,10 +969,10 @@ defmodule Req.Request do
   """
   def run_request(request)
 
-  def run_request(%{current_request_steps: [step | rest]} = request) do
-    step = Keyword.fetch!(request.request_steps, step)
+  def run_request(%{current_request_steps: [step_name | rest]} = request) do
+    step_fn = Keyword.fetch!(request.request_steps, step_name)
 
-    case step.(request) do
+    case run_step({step_name, step_fn}, request, :request) do
       %Req.Request{} = request ->
         run_request(%{request | current_request_steps: rest})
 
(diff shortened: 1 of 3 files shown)


wojtekmach:
Thank you for the PR. I’d rather not do this by default since it’s easy to do it yourself and you have more flexibility, you can add any measurement and metadata without having to configure it elsewhere. You can do:

```elixir
People can do: update_in(req.request_steps, &wrap_in_telemetry_span/1)
```

there is also small but nonzero overhead in emitting each Telemetry event so I’d rather avoid that too.

ckoch-cars:
Hi @wojtekmach. I was wondering if you would consider this req-native telemetetry approach as an opt-in option? 

I looked into emitting telemetry from CarReq and while this approach seems OK, I feel like it would be better if the telemetry can from req itself. One small reason is the masking of the step function names in the request struct itself. 

Thanks

wojtekmach:
Could you clarify masking step function names?

ckoch-cars:
> Could you clarify masking step function names?

Sure, making the change in car_req:
```
request_steps: [
    fuse: #Function<2.48640514/1 in CarReq.telemetry_spanner/2>,
    put_user_agent: #Function<2.48640514/1 in CarReq.telemetry_spanner/2>,
    compressed: #Function<2.48640514/1 in CarReq.telemetry_spanner/2>,
    encode_body: #Function<2.48640514/1 in CarReq.telemetry_spanner/2>,
```

vs in req:
```
request_steps: [
    put_user_agent: &Req.Steps.put_user_agent/1,
    compressed: &Req.Steps.compressed/1,
    encode_body: &Req.Steps.encode_body/1,
```

wojtekmach:
Got it. I could add support for steps as MFArgs:

```elixir
request_steps: [
  {Req.Steps, :put_user_agent, []},
  {CarReq, :telemetry_spanner, [&Req.Steps.compressed/1]}
]
```

Otherwise I'd reimplement Req.Request.run_request in CarReq. There's a bit of boilerplate but not too much imo. 

What do you think?

ckoch-cars:
I think either suggestion sounds good. 

I'll look into the `Req.Request.run_request` approach first, but I think the MFA support would be good too. 
I think the MFA would allow telemetry be added as a plugin, which would be nice. WDYT?

wojtekmach:
Oh, the step/plugin would basically do: `update_in(req.request_steps, &{CarReq, :telemetry_spanner, [&1]}` right? Yes that would totally work. The only problem with that however is if there are any steps _after_ this wrapper step (e.g. they have been added since) they wouldn't be wrapped.

There's basically no way to ensure that given step is pinned to be at the end. That's something I'd like to support but not sure about the API. The closest thing at the moment is `request.adapter`, which is just a step, is guaranteed to be _the_ last _request_ step.

For all these reasons I believe the best choice for you is a custom Req.Request.run_request reimplementation in CarReq.

wojtekmach:
@ckoch-cars sorry this can totally be a plugin. When we execute `update_in(req.request_steps, ...)` in a _step_ we're good. Because even if we add more steps down the road, our wrapping step is not executed until we make the request. The only concern is in theory steps can append steps and the latter wouldn't be wrapped. But that's probably an anti pattern.

wojtekmach:
I have added https://github.com/wojtekmach/req/issues/262 to the backlog.

PR #258, opened 2023-10-11
Add option to keep cookies between redirected requests

moogle19:
Not sure if this is a common case, but I had to query a server which used basic auth for the initial authentication, but then sets a cookie for the session and redirects to the wanted resource.  

This PR adds an option (`:keep_cookie`) to use the cookie received from the server via the `set-cookie` header for subsequent (redirected) requests.

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -113,7 +113,7 @@ defmodule Req do
 
     * `:url` - the request URL.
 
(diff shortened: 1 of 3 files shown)


wojtekmach:
Thanks for the PR. Do you know if other HTTP clients have something similar? This is the first time this use case is mentioned here.

moogle19:
No, not really.
Then it is probably better to not add this to `req` and I just make me a custom `redirect` step for my specific problem.

PR #259, opened 2023-10-12
Fix typespec for `put_header`

moogle19:
The typespec for `put_header/3` needs to be different between the two versions, since the new version also allows lists of binaries.


wojtekmach:
Thank you!

PR #274, opened 2023-11-19
Fix incorrect `put_header` doctest (& re-enable the doctest)

thbar:
I went through the doc during a work task and found that the `put_header` doc was still referencing "non-map" (pre 0.4) headers format.

Instead of just fixing it, I preferred to understand why the test was not failing in the first place, discovered that it was disabled, re-enabled it and fixed it.

Unsure if I missed something (re: `Req.MixProject.legacy_headers_as_lists?` in particular), or other motivations which may have lead to that specific doctest being disabled, so feel free to adapt!

### DIFF
--- a/lib/req/response.ex
+++ b/lib/req/response.ex
@@ -192,7 +192,7 @@ defmodule Req.Response do
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


wojtekmach:
Thank you!

Issue #215, opened 2023-08-10
`:zlib.inflate_nif/4` error with deflate-encoded resource


dorian-marchal:
Hello,  when fetching deflate-encoded content, `Req` fails to decode it:

```elixir
Req.get("https://httpbin.org/deflate", receive_timeout: 20_000)     
** (ErlangError) Erlang error: :data_error
    (erts 13.2.2.2) :zlib.inflate_nif(#Reference<0.3668438451.4136501253.102340>, 8192, 16384, 0)
    (erts 13.2.2.2) :zlib.dequeue_all_chunks_1/3
    (erts 13.2.2.2) :zlib.inflate/3
    (erts 13.2.2.2) :zlib.unzip/1
    (elixir 1.15.4) lib/enum.ex:2510: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir 1.15.4) lib/map.ex:916: Map.update!/3
    (req 0.3.11) lib/req/steps.ex:860: Req.Steps.decompress_body/1
    (req 0.3.11) lib/req/request.ex:755: anonymous fn/2 in Req.Request.run_response/2
    (elixir 1.15.4) lib/enum.ex:4830: Enumerable.List.reduce/3
    (elixir 1.15.4) lib/enum.ex:2564: Enum.reduce_while/3
    (req 0.3.11) lib/req/request.ex:683: Req.Request.run/1
    iex:8: (file)
```

It appears that there is a decoding issue for deflate-encoded content. `HTTPoison` seems to be able to decode it:

```elixir
HTTPoison.get("https://httpbin.org/deflate", [], recv_timeout: 20_000)
{:ok,
 %HTTPoison.Response{
   status_code: 200,
   body: <<120, 156, 61, 142, 205, 14, 130, 48, 16, 132, 239, 60, 69, 211, 179, 173, 20, 168, 46, 38, 30, 56, 24, 245,
     106, 48, 241, 138, 116, 249, 137, 210, 154, 82, 15, 74, 120, 119, 11, 36, 94, 54, 153, 111, 103, 103, 118, 8, 8,
     161, 10, 171, 103, 225, 80, 209, 29, 113, 246, 141, 43, 50, 193, 6, 11, 133, 182, 247, 108, 240, 210, 131, 147,
     233, 157, 87, 180, 113, 238, 117, 111, 53, 55, 182, 166, 179, 213, 239, 174, 61, 90, 150, 213, 168, 23, 71, 81, 62,
[...]

PR #275, opened 2023-11-24
Add support for deflate decompression

s3cur3:
EDIT: Oh! I see this was [intentionally removed](https://github.com/wojtekmach/req/issues/215#issuecomment-1675461292) in the past. Feel free to close!

This adds support for responses that were compressed using the `deflate` algorithm.

As best I can tell, this isn't terribly common for servers to send in the wild—in my own usage of Req for scraping public sites, I see gzip and Brotli way, way more frequently. However, I think it's useful to have in Req for the sake of having decompression support that matches real browsers; Safari, Firefox, and Chrome each send an `Accept-Encoding` header of `gzip, deflate, br`.

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -201,6 +201,7 @@ defmodule Req.Steps do
   Supported formats:
 
     * `gzip`
+    * `deflate`
     * `br` (if [brotli] is installed)
     * `zstd` (if [ezstd] is installed)
 
@@ -265,7 +266,7 @@ defmodule Req.Steps do
   end
 
   defp supported_accept_encoding do
-    value = "gzip"
+    value = "gzip, deflate"
     value = if brotli_loaded?(), do: "br, " <> value, else: value
     if ezstd_loaded?(), do: "zstd, " <> value, else: value
   end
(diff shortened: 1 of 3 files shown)


lib/req/steps.ex
```diff
@@ -1336,6 +1342,19 @@ defmodule Req.Steps do
...
 
+  defp decompress_body(["deflate" | rest], body, acc) do
+    # uncompress/1 is used when the body *has* zlib headers and checksum
+    # unzip/1 works when the body has no headers or checksum.
+    unzipped =
+      try do
+        :zlib.uncompress(body)
+      rescue
+        ErlangError -> :zlib.unzip(body)
+      end
```
s3cur3, line 1353: Someone with greater knowledge of the deflate representation could probably pattern match on the start of the `body` binary to determine if it had the header and checksum, but that's beyond me... 😬

test/req/steps_test.exs
```diff
@@ -704,14 +729,16 @@ defmodule Req.StepsTest do
...
+
     plug = fn conn ->
       body =
         %{a: 1}
         |> Jason.encode_to_iodata!()
         |> :zlib.compress()
 
       conn
-      |> Plug.Conn.put_resp_header("content-encoding", "deflate")
+      |> Plug.Conn.put_resp_header("content-encoding", non_existent_codec)
```
s3cur3, line 741: Now that we actually support `deflate`, we can't use it as an example of an unknown compression codec. 😄

wojtekmach:
yeah I'm happy to add it when someone actually needs but tbh I'm still unclear how exactly it should work. thanks for the PR though!

PR #285, opened 2023-12-22
Move to using the ExBrotli NIF for Brotli decompression

s3cur3:
Hi Wojtek!

This is mostly for discussion—the PR is not really ready to merge as-is for a number of reasons, but it's at least something concrete to look at.

The Erlang implementation of Brotli decompression errors on many real-world web pages I've tried to scrape with it, crashing the Req process. The backtrace looks like this:

```
** (MatchError) no match of right hand side value: {:error, :stream_failed}
    (req 0.4.5) lib/req/steps.ex:1360: Req.Steps.decompress_body/3
    (req 0.4.5) lib/req/steps.ex:1325: Req.Steps.decompress_body/1
    (req 0.4.5) lib/req/request.ex:1009: anonymous fn/2 in Req.Request.run_response/2
    (elixir 1.15.7) lib/enum.ex:4830: Enumerable.List.reduce/3
    (elixir 1.15.7) lib/enum.ex:2564: Enum.reduce_while/3
    (req 0.4.5) lib/req/request.ex:937: Req.Request.run/1
    (req 0.4.5) lib/req/steps.ex:1658: Req.Steps.redirect/1
    iex:7: (file)
```

I was able to solve my problem by using [this library](https://github.com/skygroup2/ex_brotli), which is just a NIF wrapper for Google's C implementation. On the sample corpus in the newly-added decompression test, all those pages would previously crash, but they work with the NIF. (Well... I should say with the right headers they can be _made_ to work.)

Is this something you'd be interested in taking?


mfeckie:
There's an alternative that I wrote too, which wraps rust-brotli. I chose that with Rustler for integration to avoid potential safety issues with C libraries. https://github.com/mfeckie/ex_brotli

The other thing that was helpful, imo, is using rustler_procompiled, you don't need to have all the infrastructure for compiling from source because we precompile them as a Github build step, example https://github.com/mfeckie/ex_brotli/releases/tag/0.4.0

May not solve your problem, but might be worth considering

s3cur3:
Oh, interesting! I think when last I'd looked, you hadn't added decompression yet, so I went with the C version. Yeah, Rustler precompiled is a game changer.

wojtekmach:
@s3cur3 sorry for radio silence. I basically couldn't make up my mind if we should somehow support both packages, open this up for extensibility, etc. Did you end up going with your fork or a custom step/plugin? I hope to maybe revisit it before v1.0 fwiw.

s3cur3:
Oh, no sweat! I think the right move is probably to use @mfeckie's Rust library, I just haven't had time to implement it myself. For my purposes (really good browser impersonation) I wound up having to go outside the Elixir ecosystem entirely, unfortunately, since I ended up needing to impersonate things like TLS handshakes too.

Issue #282, opened 2023-12-14
Add ability to redact custom header and option names


wojtekmach:
Currently we redact "authorization" header and :auth option. Let's make this extensible. While at it, perhaps :redact_auth should not be an _option_ but a field on the Req struct, i.e. something more fundamental.

wojtekmach:
We probably want to deprecate :redact_auth in favour of `redact_headers: ["authorization"]` and `redact_options: [:auth]` and when we override this, we always merge. In other words, there's no way to turn this off which is fine, people can always simply do `Map.take(req, [:headers, :options])` and see everything.

miguel-solano:
Just opened a WIP PR for this! Hoping to get it to a good spot within the next two weeks 👍🏽

PR #286, opened 2023-12-26
wip: Add `redact_options` & `redact_headers` support

miguel-solano:
This is an initial attempt to handle issue #282 regard redacted field support. I am currently not happy with the implementation, but wanted to open a PR to get feedback and signal that I am currently looking into this.

### DIFF
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -1013,7 +1013,7 @@ defmodule Req.Request do
 
       other ->
         raise "expected adapter to return {request, response} or {request, exception}, " <>
-                "got: #{inspect(other) }"
+                "got: #{inspect(other)}"
     end
   end
 
@@ -1101,13 +1101,11 @@ defmodule Req.Request do
       sep = color(",", :map, opts)
       close = color("}", :map, opts)
 
-      redact_options = Req.Request.get_option(request, :redact_options, [])
       redact_headers = Req.Request.get_option(request, :redact_headers, [])
 
       headers =
         if unquote(Req.MixProject.legacy_headers_as_lists?()) do
           for {name, value} <- request.headers do
-            require IEx; IEx.pry
             case Enum.member?(redact_headers, Req.__ensure_header_downcase__(name)) do
               true -> {name, "[redacted]"}
               _ -> {name, value}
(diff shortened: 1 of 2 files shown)


lib/req/request.ex
```diff
@@ -408,11 +408,15 @@ defmodule Req.Request do
         |> Keyword.update(:options, %{}, &Map.new/1)
         |> Keyword.update(
           :registered_options,
-          MapSet.new([:redact_auth]),
-          &MapSet.put(&1, :redact_auth)
-        )
+          MapSet.new([:redact_headers, :redact_options]),
+          fn registered_options ->
```
miguel-solano, line 412: Perhaps, this can be passed in as a reference with `&`. I think it would make the code a bit cleaner.

lib/req/request.ex
```diff
@@ -998,7 +1013,7 @@ defmodule Req.Request do
 
       other ->
         raise "expected adapter to return {request, response} or {request, exception}, " <>
-                "got: #{inspect(other)}"
+                "got: #{inspect(other) }"
```
miguel-solano, line 1016: whoops, accidental edit, will remove with a follow up `fixup` commit.

wojtekmach:
Thank you for the PR and apologies it has fallen through the cracks. The implementation would have to be a little bit more complicated because it would have to support nested options too, for example ideally we'd have:

```elixir
iex> Req.new(aws_sigv4: [access_key_id: "foobar", secret_access_id: "barbaz", region: "us-east-1"])
%Req.Request{
  options: %{
    aws_sigv4: [
      access_key_id: "[redacted]",
      secret_access_id: "[redacted]",
      region: "us-east-1"
    ]
  },
```

I've decided to keep the code simple and deprecate `:redact_auth`, it now has no effect, in favour of revealing the secret just a little bit to give users an idea what it is and so that they can manually take it out if they care:

```elixir
iex(1)> Req.new(auth: {:bearer, "foobar"})
%Req.Request{
  options: %{auth: {:bearer, "foo***"}},
```

and I intend to apply this for all built-in steps. And if down the road custom step authors are gonna really need this feature we can revisit.

Again apologies for no feedback on this. I really appreciate you looking into this though.

PR #293, opened 2024-01-17
Show low level Mint option usage in ReadMe

JoeARO:
When trying to disable SSL requests using Req I found it quite difficult to work out the correct usage until I eventually stumbled across the `connect_options`.
Hopefully this addition will help anyone in a similar position in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,20 @@ Req.get!(req, url: "https://api.github.com/user").body["login"]
 #=> "wojtekmach"
 ```
 
+### Using low level Mint options with Req
+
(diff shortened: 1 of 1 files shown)


wojtekmach:
Thank you for the PR and sorry for the delay. We keep Finch/Mint related docs here https://hexdocs.pm/req/Req.Steps.html#run_finch/1 and there are already some examples so I believe we're good. I'm gonna close this for now, but if think it's still not discoverable, could you send another PR that links to `run_finch/1` docs perhaps?

PR #302, opened 2024-02-16
Loosen finch version requirement

doughsay:
Is this ok to do?


wojtekmach:
Thanks, there should be a new release within a few days.

PR #303, opened 2024-02-17
fix: Use `put_header/3` instead of `put_new_header/3` in auth step.

### LINKED ISSUE #297: `Req.Steps.auth/2` does not support updating authorization header
## Problem

I am writing a `reauthenticate` step for Req that will refresh credentials on 401.

This was not working however, and I realized that the `Req.Steps.auth/2` function uses `Req.Request.put_new_header/3` and thus cannot be used for updating the authentication on a request. Looks like this needs to be done manually. It would be nice if this was supported by the `auth/2` function out of the box. The only workaround is to just copy-paste the code from the Req library.

## Proposed Solution

Have `Req.Steps.auth/2`  support updating the authorization header.

## Versions

`:req 0.4.8`

jadengis:
This change uses `put_header/3` instead of `put_new_header/3` in the `Req` auth step. This allows the auth step to be reused later in the pipeline to replace the authorization header e.g. when credentials are expired and need to be refreshed.

Closes #297.

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -179,15 +179,15 @@ defmodule Req.Steps do
   end
 
   defp auth(request, authorization) when is_binary(authorization) do
-    Req.Request.put_new_header(request, "authorization", authorization)
+    Req.Request.put_header(request, "authorization", authorization)
   end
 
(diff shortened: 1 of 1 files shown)


wojtekmach:
Thank you!

PR #314, opened 2024-02-26
Add some info to `Req.Steps.decode_body/1` docs

whatyouhide:
Formatting, plus added a link to `MIME.extensions/1` cause I had to go read the source to make sure this is what this step was doing.

cc @wojtekmach


wojtekmach:
Thank you!

PR #315, opened 2024-02-29
Add Req.Test.allow/3

whatyouhide:


lib/req/test.ex
```diff
@@ -72,6 +77,35 @@ defmodule Req.Test do
 
         assert MyApp.Weather.get_rating("Krakow, Poland") == {:ok, :nice}
       end
+
+  ### Concurrency and Allowances
```
wojtekmach, line 81: This is great.

I believe explicit allowances are not necessary a lot of the time (Task, Task.Supervisor, controller and live view test helpers propagate $callers) so I don't want to leave and impression that is always necessary when we spawn another process but I'm not sure how to document this best. Perhaps leave this vague that usually it's not necessary but if spawned process cannot find stub, it needs to be explicitly allowed.

wojtekmach (approved):

lib/req/test.ex
whatyouhide: @wojtekmach that's a really good point. Rephrased, wdyt now?

wojtekmach: Perfect!

PR #321, opened 2024-03-11
Use Req.Request.halt/1 and return redirect exception instead of raising on "too many redirects"

bfolkens:
Currently, the included `redirect` step raises an exception with a message when the maximum number of redirects is reached.  This change modifies the existing raise behavior to instead use the halt and return exception behavior used by other steps in the pipeline.


wojtekmach:
Thank you!

PR #324, opened 2024-03-16
Improve docs for Req.Test and add specs

whatyouhide:
Just a bunch of docs to clarify stuff, plus specs.

### DIFF
--- a/lib/req/test.ex
+++ b/lib/req/test.ex
@@ -165,7 +165,7 @@ defmodule Req.Test do
   like `stub/1` and `stub/2`.
   """
   @typedoc since: "0.4.15"
-  @type stub() :: atom()
+  @opaque stub() :: atom()
 
   @ownership Req.Ownership
 
--- a/lib/req/test.ex
+++ b/lib/req/test.ex
@@ -13,7 +13,9 @@ defmodule Req.Test do
     * Stub any value with [`Req.Test.stub(name, value)`](`stub/2`) and access it with
       [`Req.Test.stub(name)`](`stub/1`). These functions can be used in concurrent tests.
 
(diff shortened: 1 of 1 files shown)


whatyouhide:
cc @wojtekmach

lib/req/test.ex
```diff
@@ -154,6 +156,17 @@ defmodule Req.Test do
...
 
+  @typedoc """
+  A stub is an atom that scopes stubbed requests.
+
+  A common choice for this is the module name for the Req-based client that you
+  are using `Req.Test` for. For example, if you are using `MyApp.Weather` as
+  your client, you can use `MyApp.Weather` as the stub name in functions
+  like `stub/1` and `stub/2`.
+  """
+  @typedoc since: "0.4.15"
```
wojtekmach, line 167: It's fine to add it but fyi I'll probably remove it when I finally release v1.0.

lib/req/test.ex
```diff
@@ -154,6 +156,17 @@ defmodule Req.Test do
...
+  @type stub() :: atom()
```
wojtekmach, line 168: What do you think about changing this to a typep or an opaque? I believe in principle it can be any term and I'd like to reserve the right to change it to such if I ever need it.

wojtekmach (commented):
This is amazing.

lib/req/test.ex
whatyouhide: Yep makes sense to me 🙃

lib/req/test.ex
whatyouhide: Yep good call, done.

wojtekmach:
Thank you!

Issue #327, opened 2024-03-21
Req exits with :badarg when redirecting to invalid URL


bfolkens:
When redirect encounters an invalid URL, Req will exit with :badarg due to Finch exiting with the same.   The following error is an example:

```elixir
iex> Req.new(method: :get, url: "http://mob.muchong.com/html/201502/8544958.html") |> Req.Request.run()

 [debug] redirecting to //muchong.com/html/201502/8544958.html
** (exit) :badarg
    (finch 0.18.0) lib/finch/http1/pool.ex:96: Finch.HTTP1.Pool.request/6
    (finch 0.18.0) lib/finch.ex:472: anonymous fn/4 in Finch.request/3
    (req 0.4.13) lib/req/steps.ex:979: Req.Steps.run_finch_request/3
    (req 0.4.13) lib/req/steps.ex:837: Req.Steps.run_finch/4
    (req 0.4.13) lib/req/request.ex:993: Req.Request.run_request/1
    (req 0.4.13) lib/req/request.ex:938: Req.Request.run/1
    (req 0.4.13) lib/req/steps.ex:1977: Req.Steps.redirect/1
    (req 0.4.13) lib/req/request.ex:1010: anonymous fn/2 in Req.Request.run_response/2
    (elixir 1.15.6) lib/enum.ex:4830: Enumerable.List.reduce/3
    (elixir 1.15.6) lib/enum.ex:2564: Enum.reduce_while/3
    (req 0.4.13) lib/req/request.ex:938: Req.Request.run/1
```

The request above fails after it redirects to `//muchong.com/html/201502/8544958.html`.  As it turns out, After `URI.merge/2` in [Req.Steps.build_redirect_request/2](https://github.com/wojtekmach/req/blob/3852d7e1d0861e7c4a6ae176dba2224d1aaf786e/lib/req/steps.ex#L2005) - the URL port may become `nil` if merged with relative scheme ("//somehost.com").

```elixir
URI.merge(URI.parse("https://good.url/"), URI.parse("//test")).port == nil
```

In other words, the following triggers the failure:

```elixir
Req.new(method: :get, url: %URI{scheme: "http", host: "muchong.com", path: "/html/201502/8544958.html"}) |> Req.Request.run()
```

And the following succeeds:

```elixir
Req.new(method: :get, url: %URI{scheme: "http", port: 80, host: "muchong.com", path: "/html/201502/8544958.html"}) |> Req.Request.run()
```

[...]

PR #328, opened 2024-03-21
Normalize URI before redirect in order to avoid :badarg failure from nil port in URI

bfolkens:
Per #327 - `URI.merge/2` overwrites port attribute with nil during merge with relative redirect URI.  Since `URI.merge/2` correctly follows [RFC 3986 5.2 Relative Resolution](https://www.rfc-editor.org/rfc/rfc3986#section-5.2), a separate normalization step is needed to include [RFC 3986 6.2.3 Scheme-Based Normalization](https://www.rfc-editor.org/rfc/rfc3986#section-6.2.3).

Note: a test was not included since `Bypass` runs on a port other than 80 or 443, and we would explicitly need to test to see if the port will be defaulted to 80/443 (depending on the original URI scheme).  However, we were able to confirm a test watching for `:badarg` exit fails prior to the patch and succeeds after the patch.  This test has been excluded from the PR since it hangs the tests while waiting for a connection timeout in the success case.

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -1964,7 +1964,10 @@ defmodule Req.Steps do
           request.options[:redirect_trusted]
       end
 
-    location_url = URI.merge(request.url, URI.parse(location))
+    location_url =
+      request.url
+      |> URI.merge(URI.parse(location))
+      |> normalize_redirect_uri()
 
     request
     # assume put_params step already run so remove :params option so it's not applied again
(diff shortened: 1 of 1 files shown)


wojtekmach:
Thank you. Could you look into testing it using the :plug feature, without bypass?

bfolkens:
I tried to implement a test using the :plug feature via the following:

```elixir
test "inherit port", c do
  relative_redirect = fn
    %Plug.Conn{host: "redirected", port: nil} ->
      flunk("port cannot be nil")

    %Plug.Conn{host: "redirected", port: 80} = conn ->
      Plug.Conn.send_resp(conn, 200, "ok")

    conn ->
      conn
      |> Plug.Conn.put_resp_header("location", "//redirected")
      |> Plug.Conn.send_resp(301, "")
  end

  assert ExUnit.CaptureLog.capture_log(fn ->
           assert Req.get!(c.url, plug: relative_redirect).status == 200
         end) =~ "[debug] redirecting to //redirected"
end
```

However, in the underlying `Plug.Adapters.Test.Conn.conn/4` port is coalesced into 80 (https://github.com/elixir-plug/plug/blob/2eda50618296b1c0e5fd9d0066e382c34b28ea0a/lib/plug/adapters/test/conn.ex#L39), so it masks the failure case and the above test will always pass.

The above test at least ensures that there's no fatal error with the `normalize_redirect_uri/1` code, though I don't think it's entirely useful.  Happy to entertain any other ideas...

PR #332, opened 2024-03-25
Fix MatchError during redirect with no location header

### LINKED ISSUE #330: Redirect with missing Location header fails with MatchError
Req returns the following error when a redirect with a missing Location header is received.

```elixir
** (MatchError) no match of right hand side value: []
    (req 0.4.14) lib/req/steps.ex:1952: Req.Steps.build_redirect_request/2
    (req 0.4.14) lib/req/steps.ex:1937: Req.Steps.redirect/1
    (req 0.4.14) lib/req/request.ex:1010: anonymous fn/2 in Req.Request.run_response/2
    (elixir 1.15.6) lib/enum.ex:4830: Enumerable.List.reduce/3
    (elixir 1.15.6) lib/enum.ex:2564: Enum.reduce_while/3
    (req 0.4.14) lib/req/request.ex:938: Req.Request.run/1
```

Since RFC 2616 section 10.3 specifies that the Location header only "should" be provided (instead of "must"), redirect should handle the case where Location header is not provided.  Will submit a patch handling this case.

bfolkens:
When the redirect step is presented with a response where no `Location` header is specified, a MatchError is thrown from: https://github.com/wojtekmach/req/blob/3852d7e1d0861e7c4a6ae176dba2224d1aaf786e/lib/req/steps.ex#L1990

This patch introduces a function to retrieve and coalesce the `Location` header, avoiding a match error and respecting RFC 2616 section 10.3 (redirects).

Fixes #330

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -1932,9 +1932,11 @@ defmodule Req.Steps do
         redirect_count = Req.Request.get_private(request, :req_redirect_count, 0)
 
         if redirect_count < max_redirects do
+          location_url = resolve_redirect_location(request, response)
+
           request =
             request
-            |> build_redirect_request(response)
+            |> build_redirect_request(response, location_url)
             |> Req.Request.put_private(:req_redirect_count, redirect_count + 1)
 
           {_, result} = Req.Request.run(request)
@@ -1948,12 +1950,7 @@ defmodule Req.Steps do
     end
   end
 
-  defp build_redirect_request(request, response) do
-    [location] = Req.Response.get_header(response, "location")
-
-    log_level = Req.Request.get_option(request, :redirect_log_level, :debug)
-    log_redirect(log_level, location)
-
+  defp build_redirect_request(request, response, location_url) do
     redirect_trusted =
       case Req.Request.fetch_option(request, :location_trusted) do
         {:ok, trusted} ->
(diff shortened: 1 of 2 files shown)


wojtekmach:
Could you explain your rationale for this change?

> The user agent MAY use the Location field value for automatic redirection

(https://www.rfc-editor.org/rfc/rfc9110.html#section-15.4.2 - it obsoletes RFC 2616)

My reading of the RFC is that it doesn't say what to do when the location header is missing.

In that case I'd rather do nothing, i.e. return HTTP 303, like curl does:

```elixir
Mix.install([:bandit])

{:ok, _} =
  Bandit.start_link(
    scheme: :http,
    port: 4000,
    plug: fn
      conn, _options when conn.method == "POST" ->
        Plug.Conn.send_resp(conn, 303, "")

      conn, _options when conn.method == "GET" ->
        Plug.Conn.send_resp(conn, 200, "ok")
    end
  )

Process.sleep(:infinity)
```

```
~% curl -XPOST -L -v http://localhost:4000
*   Trying [::1]:4000...
* connect to ::1 port 4000 failed: Connection refused
*   Trying 127.0.0.1:4000...
* Connected to localhost (127.0.0.1) port 4000
> POST / HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/8.4.0
> Accept: */*
>
< HTTP/1.1 303 See Other
< date: Tue, 26 Mar 2024 11:57:57 GMT
< content-length: 0
< vary: accept-encoding
< cache-control: max-age=0, private, must-revalidate
<
* Connection #0 to host localhost left intact
```

What do you think? If you agree, could you leave just the test?

bfolkens:
Ok that makes sense - I think we should just follow the cURL behavior then above.  However, if I only leave the test then we would still encounter the bad match error when the following fails:

https://github.com/wojtekmach/req/blob/3852d7e1d0861e7c4a6ae176dba2224d1aaf786e/lib/req/steps.ex#L1990

So I've updated the branch with a reduced set of changes that handle the a more robust header match and then just return the response as-is, instead of following the request URL.  That way, the implementation follows the "MAY" condition in the RFC without a fatal error and allows the developer to handle the empty redirect.

lib/req/steps.ex
```diff
@@ -1932,13 +1932,19 @@ defmodule Req.Steps do
...
+              request =
+                request
+                |> build_redirect_request(response, location)
+                |> Req.Request.put_private(:req_redirect_count, redirect_count + 1)
+
+              {_, result} = Req.Request.run(request)
+              {Req.Request.halt(request), result}
+
+            _ ->
+              {Req.Request.halt(request), response}
```
wojtekmach, line 1946: we only want to halt when we make a new request (Req.Request.run) or raise. At the beginning of the step we have a `redirect?` conditional, basically whether this step is a no-op. Could you make the change there instead? Perhaps something like:

```
with true <- response.status in ...,
     [location | _] <- ... do
  ...
else
  _ ->
    {request, response}
end
```

bfolkens: Ok great, that's much simpler.  Also, with that in mind it looks like we could refactor up another step (to the `cond`).  I just updated branch with suggested change and also added a test for when :redirect is false.

wojtekmach:
Thank you!

Issue #335, opened 2024-03-28
Proposal: Improve error handling


wojtekmach:
Currently the contract is `Req.request(req) :: {:ok, Req.Response.t()} | {:error, Exception.t()}`, that is, users of the library might get arbitrary exceptions.

In recent releases we added a few Req exceptions:

  * `Req.TooManyRedirectsError` - previously we were _raising_ a RuntimeError, now we _return_ this exception

  * `Req.TransportError` - previously we were returning `Mint.TransportError`. We added this struct because we want to have `Req.Test.transport_error(conn, reason)` and so should have a Req specific exception.

Currently we automatically retry "transient errors". By that we mean 408/429/500/502/503/504 _responses_ and _exceptions_ with `reason: :timeout | :econnrefused | :closed`. We don't check the exception struct, just the reason.

We have a few issues:

  * #244 - should we retry all `Req.TransportError` exceptions? My rationale as not to was that for `reason: :nxdomain` it is very unlikely that retrying such would ever be successful, usually it's a typo in domain name, and so I would not consider it "transient". (IPv4/IPv6 mismatch shows up as this one too to further complicate things however.)

  * #295

  * Similarly to above, at the moment we are crashing when decoding invalid tar and zip.

  * #333

  * #329

  * Finch has [`Finch.Error`](https://hexdocs.pm/finch/Finch.Error.html) exception. Should it (or just particular :reason values) be considered _transient_?

## Proposal

  1. Add `%Req.HTTPError{reason: reason, protocol: :http1 | :http2}`. It standardizes Mint.HTTPError and Finch.Error.

  2. "transient error" means one of:

       * response 408/429/500/502/503/504,

       * `%Req.TransportError{reason: :closed | :econnrefused | :closed}`.

       * `%Req.HTTPError{reason: :unprocessed, protocol: :http2}`

  3. Remove compatibility guarantee (-ish, we're still pre 1.0 after all) for what "transient error" is, that is, document that we can add new status codes and exception structs/reasons in the future.

     If someone is really concerned about it, they'd explicitly configure `:retry` for their needs.

  4. Add `Req.DecodeError`:

     ```elixir
     %Req.DecodeError{
       format: :json | :tar | :zip | atom(), # atom() for user provided decoders
       position: non_neg_integer | nil,
       data: binary
     }
     ```

     This matches `%Jason.DecodeError{position:, data:}`. (There would be an additional private `:token` field which Jason.DecodeError has too.)

[...]

PR #337, opened 2024-03-29
Handle decompress_body/1 errors by returning Req.DecompressError

bfolkens:
See #335 - Improve decompression failure error handling by returning Req.DecompressError

This PR adds a new error type (below) to handle decompression failures.

```elixir
%Req.DecompressError{
  format: :gzip | :br | :zstd,
  data: binary()
}
```

### DIFF
--- a/lib/req/decompress_error.ex
+++ b/lib/req/decompress_error.ex
@@ -3,7 +3,12 @@ defmodule Req.DecompressError do
   Represents an error when decompression fails, returned by `Req.Steps.decompress_body/1`.
   """
 
-  defexception [:format, :data]
+  defexception [:format, :data, :reason]
+
+  @impl true
+  def message(%{format: format, reason: reason}) when not is_nil(reason) do
+    "#{message(%{format: format})}, reason: #{inspect(reason)}"
+  end
 
   @impl true
   def message(%{format: format}) do
--- a/lib/req/decompress_error.ex
+++ b/lib/req/decompress_error.ex
@@ -6,12 +6,12 @@ defmodule Req.DecompressError do
   defexception [:format, :data, :reason]
 
   @impl true
-  def message(%{format: format, reason: reason}) when not is_nil(reason) do
-    "#{message(%{format: format})}, reason: #{inspect(reason)}"
+  def message(%{format: format, reason: nil}) do
+    "#{format} decompression failed"
   end
 
   @impl true
-  def message(%{format: format}) do
-    "decompression failed with format \'#{format}\'"
+  def message(%{format: format, reason: reason}) do
+    "#{format} decompression failed, reason: #{inspect(reason)}"
   end
 end
--- a/lib/req/decompress_error.ex
+++ b/lib/req/decompress_error.ex
@@ -1,0 +1,12 @@
+defmodule Req.DecompressError do
+  @moduledoc """
+  Represents an error when decompression fails, returned by `Req.Steps.decompress_body/1`.
+  """
+
+  defexception [:format, :data]
+
+  @impl true
+  def message(%{format: format}) do
+    "decompression failed with format \'#{format}\'"
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/req/steps.ex
```diff
@@ -1636,42 +1636,70 @@ defmodule Req.Steps do
...
+         decompressed when is_binary(decompressed) <- :ezstd.decompress(body) do
+      decompress_body(rest, decompressed, acc)
     else
-      Logger.debug(":ezstd library not loaded, skipping zstd decompression")
-      decompress_body(rest, body, ["zstd" | acc])
+      false ->
+        Logger.debug(":ezstd library not loaded, skipping zstd decompression")
+        decompress_body(rest, body, ["zstd" | acc])
+
+      {:error, _reason} ->
```
bfolkens, line 1701: Do we care about the underlying error reason?  If so, we'll need to include another field in the struct...

lib/req/steps.ex
```diff
@@ -1636,42 +1636,70 @@ defmodule Req.Steps do
...
-          |> Req.Response.delete_header("content-encoding")
-          |> Req.Response.delete_header("content-length")
-        else
-          Req.Response.put_header(response, "content-encoding", Enum.join(unknown_codecs, ", "))
-        end
 
-      {request, response}
+      case decompress_body(codecs, response.body, []) do
+        %Req.DecompressError{} = exception ->
+          {Req.Request.halt(request), exception}
```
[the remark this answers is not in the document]
bfolkens: Ahh that's right - sorry about that, forgot we discussed the same during the other PR.

lib/req/decompress_error.ex
```diff
@@ -0,0 +1,17 @@
...
+
+  @impl true
+  def message(%{format: format, reason: reason}) when not is_nil(reason) do
+    "#{message(%{format: format})}, reason: #{inspect(reason)}"
+  end
+
+  @impl true
+  def message(%{format: format}) do
+    "decompression failed with format \'#{format}\'"
+  end
```
wojtekmach, line 16:
```suggestion
  @impl true
  def message(%{format: format, reason: nil}) do
    "#{format} decompression failed"
  end

  @impl true
  def message(%{format: format, reason: reason}) do
    "#{format} decompression failed, reason: #{inspect(reason)}"
  end
```

lib/req/steps.ex
```diff
@@ -1636,42 +1636,70 @@ defmodule Req.Steps do
...
       decompress_body(rest, decompressed, acc)
     else
-      Logger.debug(":brotli library not loaded, skipping brotli decompression")
-      decompress_body(rest, body, ["br" | acc])
+      false ->
+        Logger.debug(":brotli library not loaded, skipping brotli decompression")
+        decompress_body(rest, body, ["br" | acc])
+
+      :error ->
+        %Req.DecompressError{format: :br, data: body}
```
wojtekmach, line 1688: with/1 with multiple clauses tends to be hard to maintain because it isn't immediately clear where it originated from. (Right now it is but if this code grows it ain't gonna.)

Could you go with something like this?

```elixir
if brotli_loaded?() do
  case :brotli.decode(body) do
    {:ok, decompressed} ->
      # ...

    :error ->
      # ...

  end
else
  # ...
end
```

test/req/steps_test.exs
```diff
@@ -534,6 +534,18 @@ defmodule Req.StepsTest do
...
 
+    test "gzip error", c do
+      Bypass.expect(c.bypass, "GET", "/", fn conn ->
+        conn
+        |> Plug.Conn.put_resp_header("content-encoding", "x-gzip")
+        |> Plug.Conn.send_resp(200, "bad")
+      end)
+
+      assert_raise Req.DecompressError, "decompression failed with format 'gzip'", fn ->
+        Req.get!(c.url)
```
wojtekmach, line 545: nitpick: I'd like to use :plug feature as much as possible going forward, for everything that doesn't rely on network or some HTTP implementation details:

```suggestion
    test "gzip error" do
      plug = fn conn ->
        conn
        |> Plug.Conn.put_resp_header("content-encoding", "x-gzip")
        |> Plug.Conn.send_resp(200, "bad")
      end

      assert_raise Req.DecompressError, "decompression failed with format 'gzip'", fn ->
        Req.get!(plug: plug)
```

wojtekmach (commented):
Just two comments and we can ship it, thank you!

PR #339, opened 2024-03-31
fix ezstd build

MaSven:
Old ezstd tried to pull the master branch. But that does not exists anymore. Hence the build breaks with everything under ezstd < 1.0.9.


wojtekmach:
Thank you!

PR #340, opened 2024-03-31
Better Documentation for halting

### LINKED ISSUE #338: Confusing documentation regading halt signal with step functions
When using the prepend_request_steps functions in combination with Rq.Request.halt, it is quite confusing that i have to return a tuple in my step function. I think the documentation should point this out. You can kinda see it from the example, but I  was nevertheless suprised by this behaviour. 

A spec for the anonymous function that we have to prepare would here. Right now the respons for the spec would be a Request.t() or {Request.t(), Exception.t() | Response.t()}.
As the return of this anonymous function is limited by https://github.com/wojtekmach/req/blob/7616d399914eb28377d9fa7fdb0b0a0273e181cb/lib/req/request.ex#L977 would this not be a breaking change.

MaSven:
As descrbed in #338, a attempt for better documentation of the anonymous function usage in combination with halt.
Closes #338

### DIFF
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -181,6 +181,7 @@ defmodule Req.Request do
 
   Any step can call `halt/1` to halt the pipeline. This will prevent any further steps
   from being invoked.
+  You need to return a tuple with an exception or a response. Otherwise the halt is ignored.
 
   Examples:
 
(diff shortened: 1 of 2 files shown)


lib/req/request.ex
```diff
@@ -181,6 +181,7 @@ defmodule Req.Request do
 
   Any step can call `halt/1` to halt the pipeline. This will prevent any further steps
   from being invoked.
+  You need to return a tuple with an exception or a response. Otherwise the halt is ignored.
```
wojtekmach, line 184: yeah, ignoring halt was definitely a bug. I decided to deprecate halt/1 in favour of halt/2 where we're forced to return response|exception, I just pushed a commit to main.

Even though I didn't merge your PR I really appreciate your effort in general and this little note in particular as it made it obvious the previous API didn't really make sense. Thank you!

PR #346, opened 2024-04-25
Fix #345: `redirect` handler discarding the state of `Req.Request`

### LINKED ISSUE #345: `Request.private` is lost after a redirect
Hi, I noticed some odd behavior when it comes to redirects. I'm using this library with streaming responses (`into: &fun/2`) and inside that handler, I'm using `req.private` as a way to store some state that I want to retrieve after the request is completed.

However this state is being lost when there's a redirect involved. I've looked through the code and this seems to be happening because the request struct is discarded after the redirect is completed:

https://github.com/wojtekmach/req/blob/3852d7e1d0861e7c4a6ae176dba2224d1aaf786e/lib/req/steps.ex#L1973-L1979

I'd like to propose the following change (which resolves my issue), but I'm not sure if this would break any other expectations:

```diff
diff --git a/lib/req/steps.ex b/lib/req/steps.ex
index 86388bb..2053d65 100644
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -2026,7 +2026,7 @@ defmodule Req.Steps do
           |> build_redirect_request(response, location)
           |> Req.Request.put_private(:req_redirect_count, redirect_count + 1)

-        {_, result} = Req.Request.run(request)
+        {request, result} = Req.Request.run_request(request)
         Req.Request.halt(request, result)
       else
         Req.Request.halt(request, %Req.TooManyRedirectsError{max_redirects: max_redirects})
```

recursiveGecko:


wojtekmach:
Thank you!

PR #349, opened 2024-05-10
Add more Mox-like functions to Req.Test

whatyouhide:
@wojtekmach I’m not sure how you'll feel about this but to me this is step number 1 towards what we talked about. To have complete mock functionality, we'll probably need this. It's really low maintenance (Mox has been like this for the past 7-8 years?) and if we sort of vendor its code (modified to fit this) and nimble_ownership's code you'll have top-notch testing support, IMO. Thoughts?

### DIFF
--- a/lib/req/test.ex
+++ b/lib/req/test.ex
@@ -7,17 +7,17 @@ defmodule Req.Test do
   >
   > ["Mocks Aren't Stubs" by Martin Fowler](https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs)
 
-  Req already has built-in support for stubs via `:plug`, `:adapter`, and (indirectly) `:base_url`
-  options. This module enhances these capabilities by providing:
+  Req already has built-in support for different variants of stubs via `:plug`, `:adapter`,
+  and (indirectly) `:base_url` options. This module enhances these capabilities by providing:
 
-    * Stub any value with [`Req.Test.stub(name, value)`](`stub/2`) and access it with
-      [`Req.Test.stub(name)`](`stub/1`). These functions can be used in concurrent tests.
+    * Stubbing request handling with [`Req.Test.stub(name, handler)`](`stub/2`) (which can be
+      used in concurrent tests).
 
-    * Access plug stubs with `plug: {Req.Test, name}`. This works because `Req.Test` itself
-      is a plug whose job is to fetch the stubbed value under `name`. That value is
-      usually a function that takes a `Plug.Conn` and returns a `Plug.Conn`.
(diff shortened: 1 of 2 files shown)


lib/req/test.ex
```diff
@@ -167,6 +167,12 @@ defmodule Req.Test do
   @typedoc since: "0.4.15"
   @opaque stub() :: atom()
 
+  if Code.ensure_loaded?(Plug.Conn) do
+    @type plug() :: {module(), [term()]} | module() | (Plug.Conn.t() -> Plug.Conn.t())
+  else
+    @type plug() :: {module(), [term()]} | module() | (term() -> term())
```
wojtekmach, line 173: strictly speaking fun/2 is part of plug contract but fun/1 allegedly is _not_ (https://github.com/elixir-plug/plug/issues/1201#issuecomment-1872276119) so we should include it:

```suggestion
    @type plug() :: {module(), [term()]} | module() | (Plug.Conn.t() -> Plug.Conn.t() | (Plug.Conn.t(), term() -> Plug.Conn.t())
  else
    @type plug() :: {module(), [term()]} | module() | (term() -> term() | (term(), term() -> term())
```

lib/req/test.ex
```diff
@@ -273,15 +279,21 @@ defmodule Req.Test do
...
+
   @doc """
-  Returns the stub created by `stub/2`.
+  Fetches the stubbed request handler (plug) for the given `name`.
   """
-  @spec stub(stub()) :: term()
-  def stub(stub_name) do
-    case NimbleOwnership.fetch_owner(@ownership, callers(), stub_name) do
+  @doc since: "0.5.0"
+  @spec fetch_stub!(stub()) :: plug()
```
wojtekmach, line 291: wdyt about just fetch!?

```suggestion
  @spec fetch!(stub()) :: plug()
```

lib/req/test.ex
whatyouhide: Oh very good catch!!

lib/req/test.ex
whatyouhide: Mh I don't think it works. I don't think `fetch_stub!` works very well either. `Req.Test.fetch!` is like... fetching what?! Maybe `Req.Test.fetch_for_name!`? I don't think anyone is expected to use this really, so I’m not sure we need a nice name, but I’m open to other ideas as well.

wojtekmach: right, let's defp it then?

whatyouhide: Yeah let's do that for now.

wojtekmach:
Thank you!

PR #351, opened 2024-05-13
Vendor nimble_ownership

whatyouhide:
@wojtekmach as we discussed.


wojtekmach:
Thank you soo much for this!

gmile:
Out of curiosity, what prompted vendoring of nimble_ownership?

whatyouhide:
It's a test-only dep for Req, that is, only needed for testing with Req. We figured beign a pre-1.0 lib and only one file, we're better off vendoring for now. We might depend on it in the future if other non-test libs start depending on it, but for now the biggest depender on nimble_options is Mox, which is usually a test-only dep. Hope this makes sense!

PR #352, opened 2024-05-13
Add verify* functions to Req.Test

whatyouhide:
Last touch for `Req.Test`. I think this is nice because at least it completes the mock/stub interface. After this, I think we're done with `Req.Test` 🙃


wojtekmach:
Thank you!

PR #359, opened 2024-05-16
Update `Req.Test` docs

wojtekmach:
Demo:

image

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -2,13 +2,15 @@ defmodule Req do
   @moduledoc ~S"""
   The high-level API.
 
-  Req is composed of three main pieces:
+  Req is composed of:
 
     * `Req` - the high-level API (you're here!)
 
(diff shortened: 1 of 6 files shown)


wojtekmach:
@whatyouhide I went ahead and merged this but if you ever see this, any feedback as always appreciated!

I decided to use the name `name` in place of stub/mock/context/scope, etc. Also, I hid `name` and `plug` types, I really don't want to make them part of the public API. (I kinda would like to expose the type for the purpose of documenting it but not be able to refer to it from outside, if that makes sense.) Happy to change this of course if it would help.

whatyouhide:
@wojtek I agree about types not being referenced from outside but I think the docs are really nice for those. Maybe make a macro that expands to the type definition and then have 

```elixir
def stub(..., plug) when plug: unquote(type_def())
```

or something like this. I’m in Asia without a laptop so I’m just guessing around here but you get the idea 🙃

PR #362, opened 2024-05-23
Streaming response body as enumerable

wojtekmach:


test/req_test.exs
```diff
@@ -72,4 +73,29 @@ defmodule ReqTest do
...
+        {:ok, conn} = Plug.Conn.chunk(conn, "bar")
+        {:ok, conn} = Plug.Conn.chunk(conn, "baz")
+        conn
+      end)
+
+    %{url: proxy_url} =
+      start_server(fn conn ->
+        %{status: 200, body: async} = Req.get!(url: origin_url, into: :self)
+        conn = Plug.Conn.send_chunked(conn, 200)
+        Enum.reduce(async, conn, &chunk(&2, &1))
```
wojtekmach, line 91: I think this looks pretty good for a simple proxy if I may say so.

josevalim (commented):
This is fine but you need to store the process you send messages to in Req.Async and check that `self() == pid` before you start streaming. Otherwise someone can try to stream it from another process and they will just wait forever.

PR #364, opened 2024-05-30
Fix default finch protocols documentation

MrYawe:
Current code handling default protocols: https://github.com/wojtekmach/req/blob/10a3c969296de910811f77bf51af90024874634a/lib/req/steps.ex#L1065


wojtekmach:
There is a Finch bug when using H2 over H1 connections so unless that is resolved we default to using H1. Thanks for the PR though!

PR #365, opened 2024-05-31
Improve docs for Req.Request.halt/2

whatyouhide:
Tiny docs improvements + specs.

### DIFF
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -596,16 +596,22 @@ defmodule Req.Request do
   @doc """
   Halts the request pipeline preventing any further steps from executing.
 
+  This function returns an updated request and the response or exception that caused the halt.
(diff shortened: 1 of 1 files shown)


lib/req/request.ex
```diff
@@ -596,16 +596,22 @@ defmodule Req.Request do
   @doc """
   Halts the request pipeline preventing any further steps from executing.
 
+  This function returns an updated request and the response or exception that caused the halt.
+  It's perfect when used in a request step to stop the pipeline.
+
   ## Examples
 
-      def circuit_breaker(request) do
+      Req.Request.prepend_request_steps(request, circuit_breaker: fn request ->
```
whatyouhide, line 604: I think this makes it much clearer, thoughts?

wojtekmach:
Good call, thank you!

PR #366, opened 2024-06-01
Polish up Req.Request docs

whatyouhide:


lib/req/request.ex
```diff
@@ -122,45 +122,51 @@ defmodule Req.Request do
 
   Nothing is actually executed until we run the pipeline with `Req.Request.run_request/1`.
 
-  ### Request steps
+  ### Request Steps
```
whatyouhide, line 125: Consistent title casing.

lib/req/request.ex
```diff
@@ -575,14 +590,15 @@ defmodule Req.Request do
       iex> Req.Request.update_private(req, :b, 11, & &1 + 1).private
       %{a: 1, b: 11}
   """
-  @spec update_private(t(), key :: atom(), default :: term(), (atom() -> term())) :: t()
+  @spec update_private(t(), key :: atom(), default :: term(), (term() -> term())) :: t()
```
whatyouhide, line 593: This was an error I think.

lib/req/request.ex
```diff
@@ -632,6 +653,7 @@ defmodule Req.Request do
         inspect: &IO.inspect/1
       )
   """
+  @spec append_request_steps(t(), keyword(fun())) :: t()
```
whatyouhide, line 656: I think we might want to make the `request_step` and friends *public* types, as they are documented in the module doc anyway. Thoughts?

lib/req/request.ex
```diff
@@ -556,6 +570,7 @@ defmodule Req.Request do
   @doc """
   Gets the value for a specific private `key`.
   """
+  @spec get_private(t(), atom(), default) :: term() | default when default: var
```
wojtekmach, line 573: What is `var`?

wojtekmach (approved):

lib/req/request.ex
wojtekmach: Yeah we can add the type and use it in specs. I’d hold off on speccing request.request_specs though, I want to have an option to add metadata around steps in the future if I need to.

wojtekmach: Ah nevermind!

wojtekmach:
Thank you!

Issue #369, opened 2024-06-07
Would it make sense to default `:retry_log_level` to `:warning`?


jozuas:
I'm aware https://github.com/wojtekmach/req/issues/357, but while that's in the process of consideration, would it make sense to default `:retry_log_level` to `:warning`?

In my view, transient networking issues are expected, and with the default retry behaviour recoverable. What should be an error is if the retries eventually failing, not each individual retry step.

wojtekmach:
Good call, a PR would be appreciated.

PR #370, opened 2024-06-07
Default :retry_log_level to :warning

jozuas:
Closes https://github.com/wojtekmach/req/issues/369


jozuas:
I've fixed the test :)

wojtekmach:
Could you also double check docs examples? I believe we logs output in a couple of places.

jozuas:
Nice point, done.

wojtekmach:
Thank you!

PR #373, opened 2024-06-19
Add put_path_params_style/2 step

whatyouhide:
First attempt. Not sure I’m doing it right with the option + step, but it seems to work 🙃

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -412,13 +412,11 @@ defmodule Req.Steps do
 
     * `:path_params` - params to add to the templated path. Defaults to `[]`.
 
-    * `:path_params_style` (*available since v0.5.1*) - how path params are expressed.
-      Can be one of:
+    * `:path_params_style` (*available since v0.5.1*) - how path params are expressed. Can be one of:
 
-        * `:colon` (default) for Plug-style parameters, such as
-          `https://httpbin.org/status/:code`.
+         * `:colon` (default) for Plug-style parameters, such as `https://httpbin.org/status/:code`.
 
-        * `:curly` for OpenAPI-style parameters, such as `https://httpbin.org/status/{code}`.
+         * `:curly` for OpenAPI-style parameters, such as `https://httpbin.org/status/{code}`.
 
   ## Examples
 


lib/req/steps.ex
```diff
@@ -71,6 +72,7 @@ defmodule Req.Steps do
       put_base_url: &Req.Steps.put_base_url/1,
       auth: &Req.Steps.auth/1,
       put_params: &Req.Steps.put_params/1,
+      put_path_params_style: &Req.Steps.put_path_params_style/1,
```
wojtekmach, line 75: could you merge this into the put_path_params step instead? It's going to be easier to document too, just another option for that step. Finally, I don't think we need to store it in req.private and instead look at req.options when applying params. I prefer to keep as least things as possible in req.private since it's user accessible. wdyt?

whatyouhide: Suresies, incoming

whatyouhide: @wojtekmach done!

test/req/steps_test.exs
```diff
@@ -1138,7 +1148,7 @@ defmodule Req.StepsTest do
       assert_receive :ping
       refute_receive _
 
-      assert req.private == %{req_redirect_count: 3}
+      assert %{req_redirect_count: 3} = req.private
```
wojtekmach, line 1151:
```suggestion
      assert req.private == %{req_redirect_count: 3}
```

lib/req/steps.ex
```diff
@@ -398,15 +399,34 @@ defmodule Req.Steps do
...
+  digits, and underscores; this is true both for `:colon_params` as well as `{curly_params}`.
+
+  Path params are replaced in the request URL path. The path params are specified as a keyword
+  list of parameter names and values, as in the examples below. The values of the parameters are
+  converted to strings using the `String.Chars` protocol (`to_string/1`).
+
   ## Request Options
 
     * `:path_params` - params to add to the templated path. Defaults to `[]`.
+    * `:path_params_style` (*available since v0.5.1*) - how path params are expressed.
```
wojtekmach, line 414:
```suggestion

    * `:path_params_style` (*available since v0.5.1*) - how path params are expressed.
```

wojtekmach:
Thank you!

PR #374, opened 2024-06-22
CanonicalURI encoding issue in aws_sigv4

stevehodgkiss:
When using Req and the `aws_sigv4` option, AWS Bedrock endpoints that contain a `:` in the path (e.g. `https://bedrock-runtime.us-west-2.amazonaws.com/model/anthropic.claude-3-sonnet-20240229-v1:0/invoke`) fail signature verification. Model names without `:` are able to pass signature verification and run, but any with a colon fail signature verification.

```
The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.
```

I tracked it down to a path encoding issue inside `aws_sigv4_headers`. According to the [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html), the path should be mostly URI encoded. With the changes in this PR signature verification works for paths containing a `:`.

Note/warning: I have only tested these changes on this specific use case (AWS Bedrock and `Req.post(json: ...)` which uses the `aws_sigv4_headers` function in req).

### DIFF
--- a/lib/req/utils.ex
+++ b/lib/req/utils.ex
@@ -58,10 +58,7 @@ defmodule Req.Utils do
     canonical_headers =
       Enum.map_intersperse(canonical_headers, "\n", fn {name, value} -> [name, ":", value] end)
 
-    path =
-      (url.path || "/")
-      |> URI.encode(&URI.char_unreserved?/1)
-      |> String.replace("%2F", "/")
+    path = URI.encode(url.path || "/", &(&1 == ?/ or URI.char_unreserved?(&1)))
 
     canonical_request = """
     #{method}
@@ -142,10 +139,7 @@ defmodule Req.Utils do
         &String.downcase(elem(&1, 0), :ascii)
       )
 
-    path =
-      (url.path || "/")
-      |> URI.encode(&URI.char_unreserved?/1)
-      |> String.replace("%2F", "/")
+    path = URI.encode(url.path || "/", &(&1 == ?/ or URI.char_unreserved?(&1)))
 
     true = url.query in [nil, ""]
 
--- a/lib/req/utils.ex
+++ b/lib/req/utils.ex
@@ -58,9 +58,14 @@ defmodule Req.Utils do
     canonical_headers =
       Enum.map_intersperse(canonical_headers, "\n", fn {name, value} -> [name, ":", value] end)
 
+    path =
+      (url.path || "/")
+      |> URI.encode(&URI.char_unreserved?/1)
+      |> String.replace("%2F", "/")
+
     canonical_request = """
     #{method}
-    #{url.path || "/"}
+    #{path}
     #{url.query || ""}
     #{canonical_headers}
 
@@ -137,6 +142,11 @@ defmodule Req.Utils do
         &String.downcase(elem(&1, 0), :ascii)
       )
 
+    path =
+      (url.path || "/")
+      |> URI.encode(&URI.char_unreserved?/1)
+      |> String.replace("%2F", "/")
+
     true = url.query in [nil, ""]
 
     method = method |> Atom.to_string() |> String.upcase()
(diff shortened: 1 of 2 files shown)


wojtekmach:
Thank you for the PR! Could you check how aws_signature package does this and change accordingly?

wojtekmach:
I believe they have an option whether to encode URI which is on by default except when service is set to s3.

stevehodgkiss:
The encoder is implemented in the `uri_encode_path` function in aws_signature https://github.com/aws-beam/aws_signature/blob/main/src/aws_signature_utils.erl#L114-L138

AFAICT, the behaviour is the same between the Elixir version in this PR and aws_signature_utils.uri_encode_path:

```elixir
  describe "encoding differences" do
    defp encode_path(path) do
      path
      |> URI.encode(&URI.char_unreserved?/1)
      |> String.replace("%2F", "/")
    end

    test "ascii" do
      Enum.each(0..127, fn code ->
        char = <<code>>
        assert :aws_signature_utils.uri_encode_path(char) == encode_path(char)
      end)
    end

    defp valid_codepoint?(codepoint) do
      codepoint not in 0xD800..0xDFFF
    end

    test "unicode chars" do
      0x0000..0x10FFFF
      |> Enum.filter(&valid_codepoint?/1)
      |> Enum.each(fn codepoint ->
        char = <<codepoint::utf8>>

        assert :aws_signature_utils.uri_encode_path(char) == encode_path(char)
      end)
    end
  end
```

> I believe they have an option whether to encode URI which is on by default except when service is set to s3.

It appears to be on by default, but the wording sounds like it might not be strictly required by S3: https://github.com/aws-beam/aws_signature/blob/main/src/aws_signature.erl#L46-L50

Had a quick test with S3. The following works on this PR, but fails on `main` due to the signature verification issue.

```elixir
s3_bucket = "a-bucket-name"
s3_region = "us-west-2"

req =
  Req.new(
    base_url: "https://#{s3_bucket}.s3-#{s3_region}.amazonaws.com/",
    aws_sigv4: [
      access_key_id: System.get_env("AWS_ACCESS_KEY_ID"),
      secret_access_key: System.get_env("AWS_SECRET_ACCESS_KEY"),
      region: s3_region,
      service: :s3
    ]
  )

key = "/key:1"
%{status: 200} = Req.put!(req, url: key, body: "Hello, World!")
%{status: 200, body: "Hello, World!"} = Req.get!(req, url: key)
```

lib/req/utils.ex
```diff
@@ -58,9 +58,14 @@ defmodule Req.Utils do
     canonical_headers =
       Enum.map_intersperse(canonical_headers, "\n", fn {name, value} -> [name, ":", value] end)
 
+    path =
+      (url.path || "/")
+      |> URI.encode(&URI.char_unreserved?/1)
+      |> String.replace("%2F", "/")
```
wojtekmach, line 64:
```suggestion
    path = URI.encode(url.path || "/", & &1 == ?/ or URI.char_unreserved?(&1))
```

test/req/utils_test.exs
```diff
@@ -15,7 +15,7 @@ defmodule Req.UtilsTest do
         service: "s3",
         datetime: ~U[2024-01-01 09:00:00Z],
         method: :get,
-        url: "https://s3",
+        url: "https://s3/test/path:v1/invoke",
```
wojtekmach, line 18: nitpick: let's use more generic path? this one seemed related to some kind of api other than s3 maybe?

```suggestion
        url: "https://s3/foo/:bar",
```

wojtekmach (commented):
Thanks for additional context. This looks good to me, I left couple minor comments below and it's good to go.

wojtekmach:
Thank you!

PR #376, opened 2024-06-24
Fix spec on register_options/2

notslang:
The code example on this function is this:

```elixir
|> Req.Request.register_options([:foo])
```

Which shows a list of atoms being passed in, so I don't think `keyword()` is the right spec.

### DIFF
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -524,7 +524,7 @@ defmodule Req.Request do
       iex> Req.Request.get_option_lazy(req, :b, fun)
       42
   """
-  @spec get_option_lazy(t(), atom(), (-> term())) :: term()
+  @spec get_option_lazy(t(), atom(), (() -> term())) :: term()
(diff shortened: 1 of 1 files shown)


lib/req/request.ex
```diff
@@ -524,7 +524,7 @@ defmodule Req.Request do
       iex> Req.Request.get_option_lazy(req, :b, fun)
       42
   """
-  @spec get_option_lazy(t(), atom(), (-> term())) :: term()
+  @spec get_option_lazy(t(), atom(), (() -> term())) :: term()
```
notslang, line 527: got this from running mix format

wojtekmach, line 527:
```suggestion
  @spec get_option_lazy(t(), atom(), (-> term())) :: term()
```

wojtekmach: this is because you're probably using older Elixir version. We always use the latest for formatting!

wojtekmach:
Thank you!

PR #381, opened 2024-07-04
Add support for Zstandard content-type

dlindenkreuz:
Although Req has long supported Zstandard `content-encoding`, it does not yet support Zstandard `content-type` which is used for original `.zst` archive files (as opposed to encoding other original media types like JSON for transfer).

This PR adds automatic decompression when `:ezstd` is available and `content-type: application/zstd` is detected. It depends on an upstream change in the MIME package: https://github.com/elixir-plug/mime/pull/82


wojtekmach:
Thanks for the PR, please add docs and test and this is ready to go!

lib/req/steps.ex
```diff
@@ -1569,6 +1569,20 @@ defmodule Req.Steps do
...
   end
 
+  defp decode_body({request, response}, "zst") do
+    if ezstd_loaded?() do
+      case :ezstd.decompress(response.body) do
+        decompressed when is_binary(decompressed) ->
+          {request, put_in(response.body, decompressed)}
+
+        {:error, _} ->
+          {request, %Req.ArchiveError{format: :zstd, data: response.body}}
```
wojtekmach, line 1579: zstd is not an archive like tar and zip though, I don't think ArchiveError is appropriate here. please return a RuntimeError with custom message.

dlindenkreuz:
Done. In the meanwhile, MIME has been updated to 2.0.6, which adds support for Zstd.

Req currently requires MIME `~> 1.6 or ~> 2.0` — should we leave a note somewhere that MIME `~> 2.0.6` is required for Zstd?

test/req/steps_test.exs
```diff
@@ -870,6 +870,40 @@ defmodule Req.StepsTest do
...
+
+    test "zstd invalid" do
+      plug = fn conn ->
+        conn
+        |> Plug.Conn.put_resp_content_type("application/zstd", nil)
+        |> Plug.Conn.send_resp(200, "bad")
+      end
+
+      assert {:error, e} = Req.get(plug: plug)
+      assert match?(%RuntimeError{}, e)
```
wojtekmach, line 901:
```suggestion
      assert %RuntimeError{} = e
```

wojtekmach:
Thank you!

PR #391, opened 2024-07-18
use `:finch` value to override finch instance

slapers:
Hi, 
I was unable to override the finch instance and traced it back to commit https://github.com/wojtekmach/req/commit/7ee7fd8effe22fd59d87723c86ba652d4a0c6a0e which removed the functionality.

This PR should revert that + adds a test to check that the `:finch` option can be used.


wojtekmach:
Oof, good catch, thank you for the PR. We should have a new release by end of the week.

PR #396, opened 2024-07-24
Make docs a badge link

PragTob:
Maybe it's just me preferring the badges but the docs link felt a bit lost on its own and it seems better with the badges up top.

Feel free to close if you disagree :green_heart:


wojtekmach:
Thanks, I went with `documentation-gray` that Phoenix & Ecto also use!

PR #415, opened 2024-09-26
make the README runnable in Livebook :)

spencerolson:


spencerolson:
took surprisingly few changes to get this working in Livebook! There is one thing I still can't figure out: with the final code code sample in the README i get

```
** (Mix.Error) Mix.install/2 can only be called with the same dependencies in the given VM
    (mix 1.15.7) lib/mix.ex:577: Mix.raise/2
    #cell:dy5ab6jj3lj5ooey:1: (file)
```

And I'm not sure of any way around that other than requiring all of those dependencies at the top in setup (and I felt that'd mess with the flow and intent of the README to have those all installed at the top).

README.livemd
```diff
@@ -0,0 +1,235 @@
+# Req
+
+## Section
+
+[![CI](https://github.com/wojtekmach/req/actions/workflows/ci.yml/badge.svg)](https://github.com/wojtekmach/req/actions/workflows/ci.yml)
+[![License](https://img.shields.io/hexpm/l/req.svg)](https://github.com/wojtekmach/req/blob/main/LICENSE.md)
+[![Version](https://img.shields.io/hexpm/v/req.svg)](https://hex.pm/packages/req)
+[![Hex Docs](https://img.shields.io/badge/documentation-gray.svg)](https://hexdocs.pm/req)
+
+[![Run in Livebook](https://livebook.dev/badge/v1/blue.svg)](https://livebook.dev/run?url=https%3A%2F%2Fgithub.com%2Fwojtekmach%2Freq%2Fblob%2Fmain%2FREADME.livemd)
```
spencerolson, line 10: Adds a "Run in Livebook" button.

The rest of the changes are minor changes to the code samples to make them work in Livebook.

spencerolson:
Check it out in action!

https://github.com/user-attachments/assets/b5cbc667-c785-4b8d-9f7d-8209e3ee6bc1

wojtekmach:
This is a really nice idea! We cannot remove README.md though as it's rendered on GitHub which is important to me. How about we make any necessary changes to README.md and have a README.livemd which is the exact copy (we can have mix.exs keep them identical)?

> ```
> ** (Mix.Error) Mix.install/2 can only be called with the same dependencies in the given VM
> ```

Yeah not sure what to do about that either maybe we don't have an exact copy after all and we skip this section in the README.livemd.

spencerolson:
Given that GitHub will render `.livemd` files as markdown, is this something you would consider merging in @wojtekmach ? Thanks!

wojtekmach:
I was unsure whether README.livemd is considered _the_ README, i.e. rendered when going to project page, and it sure does. So I think this is viable, I'll think about for sure, thank you!

PR #417, opened 2024-10-04
Bearer token can be fetched dynamically.

chgeuer:
Hi @wojtekmach,

This PR allows the bearer token to be fetched dynamically, by supplying a 0-arity function:

```elixir
Req.new(
   auth: {:bearer, fn -> "..." end}
)
```

This is e.g. helpful when having a GenServer (like [here](https://github.com/thmsmlr/instructor_ex/blob/main/pages/azure-openai.md#microsoft-entra-id-authentication)) that constantly keeps a valid token in-memory. This way, you just tell :req to grab the latest token whenever it needs to make a request.

I haven't open an issuer for this, can do so if you'd prefer that.

Without a bearer token being dynamically fetched, I do something like this (which is like, nah):

```elixir
defmodule SomeClientWithDynamicBearerFetching do
  @new_opts NimbleOptions.new!(bearer_token_func: [type: {:fun, 0}, required: true])

  def new(req, opts) do
    %{
      bearer_token_func: bearer_token_func
    } = opts |> opts_to_map(@new_opts)

    req
    |> attach_dynamic_bearer_token(bearer_token_func)
  end

  def new(opts \\ []) do
    %{
      bearer_token_func: bearer_token_func
    } = opts |> opts_to_map(@new_opts)

    Req.new()
    |> attach_dynamic_bearer_token(bearer_token_func)
  end

  def attach_dynamic_bearer_token(req, bearer_token_func)
      when is_function(bearer_token_func, 0) do
    req
    |> Req.Request.put_private(:bearer_token_func, bearer_token_func)
    |> Req.Request.prepend_request_steps(dynamically_add_auth_bearer: &add_fresh_bearer/1)
  end

  defp add_fresh_bearer(req) do
    case Req.Request.get_private(req, :bearer_token_func) do
      func when is_function(func, 0) ->
        req
        |> Req.Request.merge_options(auth: {:bearer, func.()})

      nil ->
        req
    end
  end

  def opts_to_map(opts, nimble_options_schema) do
    {:ok, opts} = NimbleOptions.validate(opts, nimble_options_schema)
    Map.new(opts)
  end
end
```


wojtekmach:
PR is much appreciated, thank you!

I'd like to take some time thinking this through. If we allow `auth: {:bearer, fun}` do we allow `auth: {:basic, fun}` too? It doesn't seem that useful however I think there's an argument to be made to make things uniform.

FWIW we can already achieve this today like this:

```elixir
Req.new()
|> Req.Request.prepend_request_steps(
  auth_bearer_dynamic: fn req ->
    with %{options: %{auth: {:bearer, fun}}} when is_function(fun) <- req do
      put_in(req.options[:auth], {:bearer, fun.()})
    end
  end
)
```

chgeuer:
In my opinion, the dynamic capability is not so important for basic auth (unless you want symmetry). A basic authN is certainly a known user password, which doesn't change or expire frequently. However, a bearer token typically expires after a (short) period of time.

wojtekmach:
Right, sorry, I meant to say for basic auth dynamic function seems unnecessary but we may need to add it for consistency if we were add it for bearer auth. But maybe we don't. I'll think about it some more. Thank you.

wojtekmach:
@chgeuer OK, I think I have the answer, a more generalised solution without explicitly supporting `{:basic, fun}`, instead of:

```elixir
auth: {:bearer, fn -> ... end}
```

let's go with:

```elixir
auth: fn -> {:bearer, ...} end
```

in other words instead of:

```elixir
@type auth :: string | {:basic, string} | {:bearer, string}
```

we have:

```elixir
@type auth :: auth | (-> auth)
      when auth:
             string
             | {:basic, string}
             | {:bearer, string}
```

WDYT?

If this sounds good, please remember to update "redact" test in test/req_test.exs.

chgeuer:
@wojtekmach Can we run this in the CI/CD here?

lib/req.ex
```diff
@@ -223,6 +223,8 @@ defmodule Req do
 
         * `string` - sets to this value.
 
+        * `&auth_generator/1` - a function that returns one of the above (such as a `{:bearer, token}`).
```
wojtekmach, line 226:
```suggestion
        * `&fun/0` - a function that returns one of the above (such as a `{:bearer, token}`).
```

lib/req/request.ex
```diff
@@ -1256,6 +1256,10 @@ defmodule Req.Request do
       {:bearer, redact(bearer)}
     end
 
+    defp redact_option(:auth, auth_generator) when is_function(auth_generator, 0) do
+      auth_generator
+    end
```
wojtekmach, line 1261:
```suggestion
    defp redact_option(:auth, fun) when is_function(fun, 0) do
      fun
    end
```

lib/req/steps.ex
```diff
@@ -224,6 +232,16 @@ defmodule Req.Steps do
...
 
+  defp auth(request, auth_generator) when is_function(auth_generator, 0) do
+    generated_auth_val = auth_generator.()
+
+    if is_function(generated_auth_val , 0) do
+      IO.warn("setting `auth: fn -> ... end` should not return another function; ")
+    end
+
+    auth(request, generated_auth_val)
+  end
```
wojtekmach, line 243:
```suggestion
  defp auth(request, fun) when is_function(fun, 0) do
    value = fun.()

    if is_function(value, 0) do
      raise ArgumentError, "setting `auth: fn -> ... end` should not return another function"
    end

    auth(request, value)
  end
```

wojtekmach (approved):

wojtekmach:
Thank you!

chgeuer:
Thank *you* for creating and maintaining it, good stuff.

chgeuer:
QQ, would it be possible to release this as 0.5.8 to hex.pm?

PR #433, opened 2024-11-20
fix: encode path in url for aws_sigv4_url/1

fahchen:
I encountered a `SignatureDoesNotMatch` when the object key contains characters that need to be escaped, such as spaces or CJK characters.

The `path` has been encoded at 168L, so we just update the `url` before returning.
https://github.com/wojtekmach/req/blob/a3c48b845119069909e9197cd53264782513b4b4/lib/req/utils.ex#L168

### DIFF
--- a/lib/req/utils.ex
+++ b/lib/req/utils.ex
@@ -201,7 +201,7 @@ defmodule Req.Utils do
         secret_access_key
       )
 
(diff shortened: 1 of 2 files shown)


wojtekmach:
Thank you!

PR #439, opened 2024-12-03
Clarify that raw disables both decompression and decoding

PragTob:
Reading the docs for decoding I was confused like "Wait, what is the difference between `:raw` and `:decode_body` then?" And well that is it, searching for `:raw` more unveils it but I think since the option affects both steps mentioning both effects may be worth it :)

Thanks for your excellent work on req! :green_heart:

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -1390,7 +1390,7 @@ defmodule Req.Steps do
 
   ## Options
 
-    * `:raw` - if set to `true`, disables response body decompression. Defaults to `false`.
+    * `:raw` - if set to `true`, disables response body decompression & decoding. Defaults to `false`.
 
   ## Examples
 
@@ -1582,7 +1582,7 @@ defmodule Req.Steps do
 
     * `:decode_json` - options to pass to `Jason.decode/2`, defaults to `[]`.
 
-    * `:raw` - if set to `true`, disables response body decoding. Defaults to `false`.
+    * `:raw` - if set to `true`, disables response body decoding & decompression. Defaults to `false`.
 
   ## Examples
 


PragTob:
IMG_20171227_094933_Bokeh

lib/req/steps.ex
```diff
@@ -1390,7 +1390,7 @@ defmodule Req.Steps do
 
   ## Options
 
-    * `:raw` - if set to `true`, disables response body decompression. Defaults to `false`.
+    * `:raw` - if set to `true`, disables response body decompression & decoding. Defaults to `false`.
```
wojtekmach, line 1393: This is fine but wdyt about something even more explicit, along the lines of?

```suggestion
    * `:raw` - if set to `true`, disables response body decompression. Defaults to `false`.
    
      Note: `:raw` option also disables response body decoding in the `decode_body/1` step.
```

PragTob: Sounds great to me! I didn't wanna be as intrusive as I thought there may have been a reason for not mentioning it. Let me fix that up :)

lib/req/steps.ex
```diff
@@ -1582,7 +1584,7 @@ defmodule Req.Steps do
 
     * `:decode_json` - options to pass to `Jason.decode/2`, defaults to `[]`.
 
-    * `:raw` - if set to `true`, disables response body decoding. Defaults to `false`.
+    * `:raw` - if set to `true`, disables response body decoding & decompression. Defaults to `false`.
```
wojtekmach, line 1587:
```suggestion
    * `:raw` - if set to `true`, disables response body decoding. Defaults to `false`.

    Note: `:raw` option also disables response body decompression in the `decompress_body/1` step.
```

wojtekmach:
Thank you!

PragTob:
lmao you were faster than me I just updated the other bit :grin:

Issue #386, opened 2024-07-16
`json`: Support optional use of `:json` in OTP 27+


zachallaun:
With the release of `:json` in OTP 27, I think it's worth discussing how Req might support its (optional) use. I don't feel strongly that Req _should_ support it (more on that below), but I figured a discussion would be worth having that could at least be pointed to in the future when the topic comes up!

Some points for context:

- As of v0.5.2, Jason is one of Req's three required dependencies, alongside Finch and MIME. (Other de/encoding deps, such as NimbleCSV, brotli, et al. are optional.)
- Req uses the following from Jason, not all of which have a direct correspondence in `:json`:
  - [`Jason.encode_to_iodata!/1`](https://hexdocs.pm/jason/Jason.html#encode_to_iodata!/2) (opts not supported) - corresponds to [`:json.encode/1`](https://erlang.org/documentation/doc-15.0-rc3/lib/stdlib-6.0/doc/html/json.html#encode/1) (returns iodata and raises on error)
  - [`Jason.encode!/1`](https://hexdocs.pm/jason/Jason.html#encode!/2) (opts not supported) - equivalent to `value |> :json.encode() |> IO.iodata_to_binary()`
  - [`Jason.decode/2`](https://hexdocs.pm/jason/Jason.html#decode/2) (opts supported with `:decode_json`) - no direct correspondence, [`:json.decode/3`](https://erlang.org/documentation/doc-15.0-rc3/lib/stdlib-6.0/doc/html/json.html#decode/3) raises on error and does not support high-level opts like `keys: :atoms`, but requires that the same effect be implemented using callbacks
- One potentially critical component missing from `:json` is the protocol support that Jason provides for struct encoding. Whereas Jason will error when attempting to encode a struct that doesn't implement `Jason.Encoder`, `:json` will happily encode to `{"__struct__": "Elixir.MyStruct", ...}`.

Given that `:json` is not a drop-in replacement for Jason, I see a few ways forward:

1. Do nothing and wait. There's an issue in the Jason repo ([michalmuskala/jason#185](https://github.com/michalmuskala/jason/issues/185)) suggesting that the current plan is to propose an Elixir standard library wrapper around `:json`. We could wait for that and continue to depend on `Jason`. (It's possible that this wrapper, if accepted, would not be implemented until Elixir drops support for OTP 26, which I believe is 1.19 at the earliest.)
[...]

PR #441, opened 2024-12-24
[DRAFT] use Elixir.JSON rather than Jason for elixir 18.0+

joshua-bouv:
This PR isn't quite complete as 1 test is still failing

```
  1) test compress_body request (Req.StepsTest)
     test/req/steps_test.exs:327
     ** (FunctionClauseError) no function clause matching in JSON.decode!/1

     The following arguments were given to JSON.decode!/1:

         # 1
         [~c"{", [[34, "a", 34], 58 | "1"], [], ~c"}"]

     Attempted function clauses (showing 1 out of 1):

         def decode!(binary) when is_binary(binary)

     code: assert @json.decode!(req.body) == %{"a" => 1}
     stacktrace:
       (elixir 1.18.1) lib/json.ex:307: JSON.decode!/1
       test/req/steps_test.exs:329: (test)
```

I will take another look at this tomorrow but id be keen to hear some opinions on this. I assume it's because the elixir JSON module doesn't support converting charlists (where Jason previously has). Hoping this won't be a major blocker...  🤞

Also this is my first PR of this kind, so hopefully I've done a decent job! Please tell me off for any naughty's!!

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -264,7 +264,7 @@ defmodule Req do
     * `:decode_body` - if set to `false`, disables automatic response body decoding.
       Defaults to `true`.
 
-    * `:decode_json` - options to pass to `Jason.decode!/2`, defaults to `[]`.
+    * `:decode_json` - options to pass to `JSON.decode!/2`, defaults to `[]`.
 
     * `:into` - where to send the response body. It can be one of:
 
--- a/lib/req/request.ex
+++ b/lib/req/request.ex
@@ -179,7 +179,7 @@ defmodule Req.Request do
       def decode({request, response}) do
         case Req.Response.get_header(response, "content-type") do
           ["application/json" <> _] ->
-            {request, update_in(response.body, &Jason.decode!/1)}
+            {request, update_in(response.body, &JSON.decode!/1)}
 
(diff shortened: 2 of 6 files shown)


joshua-bouv:
On reflection, as the `json_options` args are not interchangeable between Jason and JSON, maybe instead a different approach of a preferred json library should be passed to `req` so you explicitly know which one you are using and is not determined by your elixir version, which bumping could break everything everything.

wojtekmach:
Thank you for the PR. Note, this is a breaking change for `json_options` you mentioned as well as because people might be relying on Jason.Encoder protocol implementations. I haven't made up my mind yet, I'll either continue to depend on :jason, or make it optional, or for Req v1.0 require Elixir v1.18. I've been for a while thinking about passing custom decoders, e.g.: `decoders: [csv: NimbleCSV.RFC4180.parse_string/1]` and so there could be a corresponding `encoders: [json: &Jason.encode_to_iodata(&1, pretty: true)]`.

I'm going to close this in favor of the existing issue, https://github.com/wojtekmach/req/issues/386, but please feel free to continue discussing it. Thanks again!

PR #443, opened 2025-01-02
Support custom headers in aws_sigv4_url/1

benjreinhart:
S3-compatible APIs support additional headers for certain behaviors. For example, my team wants to create presigned urls for uploading files from the client. To ensure integrity of uploads and keep track of storage consumption by our end users, we want to enforce the uploaded file is the size reported by the client. This can be accomplished using the `content-length` header.

S3 supports `content-length-range` which I see is used by req_s3 `presign_form/1`. However, we're using Cloudflare R2 via their S3-compatible API and it appears that value is not supported but the `content-length` header is. Additionally, we want the exact size to be enforced, not a range or max.

This PR supports arbitrary headers passed to `Req.Utils.aws_sigv4_url/1`. I wasn't entirely sure where this belongs as there is S3-signing-related logic in a few places (this module and req_s3). From what I could tell, signature logic should be added here. Happy to adjust if needed.

### DIFF
--- a/lib/req/utils.ex
+++ b/lib/req/utils.ex
@@ -139,6 +139,7 @@ defmodule Req.Utils do
     {method, options} = Keyword.pop!(options, :method)
     {url, options} = Keyword.pop!(options, :url)
     {expires, options} = Keyword.pop(options, :expires, 86400)
+    {headers, options} = Keyword.pop(options, :headers, [])
     [] = options
 
     datetime = DateTime.truncate(datetime, :second)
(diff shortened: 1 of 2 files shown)


lib/req/utils.ex
```diff
@@ -147,22 +148,25 @@ defmodule Req.Utils do
...
+      URI.encode_query(
+        [
+          {"X-Amz-Algorithm", "AWS4-HMAC-SHA256"},
+          {"X-Amz-Credential",
+           "#{access_key_id}/#{date_string}/#{region}/#{service}/aws4_request"},
+          {"X-Amz-Date", datetime_string},
+          {"X-Amz-Expires", expires},
+          {"X-Amz-SignedHeaders", signed_headers}
+        ],
+        # Ensure spaces are encoded as %20 not +
```
benjreinhart, line 168: While this should not be a problem with the existing implementation (because arbitrary query params are not supported), this felt like a potential bug waiting to happen, so I updated to use the correct whitespace encoding.

wojtekmach:
Thank you!

PR #444, opened 2025-01-05
Support any enumerable in mutlipart

### LINKED ISSUE #442: Multipart should support streaming bodies
Per https://hexdocs.pm/req/Req.Steps.html#encode_body/1 
`:form_multipart` supports only
- integer
- iodata
- File.Stream

So, this is not possible as of today.
```elixir
%{body: body} = Req.get!("...", into: self) # todo: back pressure
file = {body, content_type: "video/mp4", filename: "..."} # todo: include :size
Req.post!(form_multipart: [file: file])
```

I tried to patch the lib and surprisingly it worked. See https://github.com/wojtekmach/req/compare/main...knightpp:req:add-multipart-stream

My use case is: restream video bytes from a GET request into POST multipart.

knightpp:
Closes #442

### DIFF
--- a/lib/req/utils.ex
+++ b/lib/req/utils.ex
@@ -464,6 +464,11 @@ defmodule Req.Utils do
     }
   end
 
+  defp add_form_parts({_parts1, size1}, {_parts2, size2})
+       when not is_integer(size1) or not is_integer(size2) do
+    raise ArgumentError, "multipart part sizes must be integers"
+  end
+
   defp add_form_parts({parts1, size1}, {parts2, size2})
        when is_list(parts1) and is_list(parts2) do
     {[parts1, parts2], size1 + size2}
@@ -474,7 +479,7 @@ defmodule Req.Utils do
   end
 
   defp encode_form_part({name, {value, options}}, boundary) do
-    options = Keyword.validate!(options, [:filename, :content_type])
+    options = Keyword.validate!(options, [:filename, :content_type, :content_length])
 
     {parts, parts_size, options} =
       case value do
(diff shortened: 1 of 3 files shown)


knightpp:
@wojtekmach is there anything I can do to fix failing tests? I see one CI job passed.

wojtekmach:
Thank you!

lib/req/utils.ex
```diff
@@ -504,6 +509,12 @@ defmodule Req.Utils do
             end)
 
           {stream, size, options}
+
+        enum ->
+          Enumerable.impl_for!(enum)
+          size = Keyword.get(options, :content_length, 0)
```
wojtekmach, line 515: @knightpp oh, I don't think `0` is a good default:

```
iex> Req.post!("https://httpbin.org/anything", form_multipart: [file: {Stream.cycle(["abc"]) |> Stream.take(3), filename: "a.txt"}]).body["files"]
%{}
iex> Req.post!("https://httpbin.org/anything", form_multipart: [file: {Stream.cycle(["abc"]) |> Stream.take(3), filename: "a.txt", content_length: 9}]).body["files"]
%{"file" => "abcabcabc"}
```

should I change this to `size = Keyword.fetch!(options, :content_length)`?

test/req/utils_test.exs
```diff
@@ -175,6 +175,45 @@ defmodule Req.UtilsTest do
...
+    test "can accept any enumerable" do
+      enum = Stream.cycle(["a"]) |> Stream.take(1)
+
+      %{body: body, size: size} =
+        Req.Utils.encode_form_multipart([field1: {enum, content_length: 11}], boundary: "foo")
+
+      body = body |> Enum.to_list() |> IO.iodata_to_binary()
+
+      # content_length is +10 of actual size
+      assert size == byte_size(body) + 10
```
wojtekmach, line 214: FYI we should have kept this property from the above test:

```
assert size == byte_size(body)
```

i.e. the `size` will be used as the `content-length` of the entire HTTP request and the `body` will be that http request body`.

Issue #342, opened 2024-04-09
JSON returned, but not parsed as JSON


Arsenalist:
I'm calling a few Facebook APIs and even though JSON is returned from the APIs, I'm forced to manually parse it using Jason.decode() instead of Req doing it automatically.  Basically, the value is a string of JSON instead of a Map.

`result = Req.post!("https://graph.facebook.com...").body`

This doesn't happen on all calls but some. Is there a way I can force it to decode as JSON?

wojtekmach:
what is the content-type of responses that aren’t automatically decoded?

you can add a response step that always decodes:

```elixir
Req.new(decode_body: false)
|> Req.Request.append_response_steps(
  decode_json: fn req, resp ->
    {req, update_in(resp.body, &Jason.decode!/1}
  end
)
```

wojtekmach:
ping

Arsenalist:
> what is the content-type of responses that aren’t automatically decoded?
> 
> you can add a response step that always decodes:
> 
> ```elixir
> Req.new(decode_body: false)
> |> Req.Request.append_response_steps(
>   decode_json: fn req, resp ->
>     {req, update_in(resp.body, &Jason.decode!/1}
>   end
> )
> ```

Sorry about the delay. Thank you for this approach. I'm using Meta's Marketing API and though JSON is returned, it doesn't seem to parse it. The content-header is `application/json`. However, doing what you suggested works fine. Thank you.

caioaao:
I'm also experiencing this, in version 0.5.8. I also checked that the content type is application/json (more specifically `"content-type" => ["application/json; charset=utf-8"]`, and only happens once every 30 or so requests

wojtekmach:
@caioaao can you share anything else about the response? What's the status code? Did it happen after a redirect or a retry?

caioaao:
Sure! This is most of the response (omitting part of the body since it's a production payload):

```elixir
 %Req.Response{
    status: 200,
    headers: %{
      "cache-control" => ["max-age=0, no-cache, no-store"],
      "connection" => ["keep-alive"],
      "content-length" => ["2518"],
      "content-type" => ["application/json; charset=utf-8"],
      "date" => ["Thu, 16 Jan 2025 21:03:57 GMT"],
      "expires" => ["Thu, 16 Jan 2025 21:03:57 GMT"],
      "pragma" => ["no-cache"],
      "server" => ["nginx"],
      "x-client-tls-ver" => ["tls1.3"],
    },
[...]

PR #450, opened 2025-01-17
Do not carry `halt` status between retries

caioaao:
When retrying a request that halted, but the halt reason is retriable, the `halted: true` shouldn't carry over to the next retry.

Fixes https://github.com/wojtekmach/req/issues/342


caioaao:
@wojtekmach did you have a chance to look into this?

wojtekmach:
Thank you!

PR #445, opened 2025-01-05
Support custom query params in Req.Utils.aws_sigv4_url/1

benjreinhart:
Similar to https://github.com/wojtekmach/req/pull/443, S3-compatible APIs support additional query parameters for certain behaviors. One example is to set the response content disposition for file downloads, as motivated by https://github.com/wojtekmach/req_s3/issues/8.

This PR supports arbitrary query parameters passed to `Req.Utils.aws_sigv4_url/1`.

PR #455, opened 2025-02-14
`:query_params` option for `aws_sigv4_url/1`

balexand:
This adds support for adding query params to presigned URLs. S3 allows response headers to be overridden using query params such as `response-cache-control`, `response-content-disposition`, `response-content-encoding`, etc. See https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html for details.

I've tested this implementation by making requests to S3 with `response-content-disposition` and `response-cache-control`.

If this gets merged then I'll submit a separate pull-request to add support for this to `ReqS3.presign_url/1`.

Thanks!


balexand:
This is a duplicate of #445

PR #457, opened 2025-02-16
Support any enumerable in mutlipart 2

### LINKED ISSUE #456: Multipart should support streaming bodies 2
Continuation of https://github.com/wojtekmach/req/issues/442

knightpp:
Closes #456 

- adds `:content_length` (maybe rename to `:size`?)
- the new option IS required when value type is `Enumerable`
- no validation :( users can craft requests with incorrect content-length header

### DIFF
--- a/test/req/utils_test.exs
+++ b/test/req/utils_test.exs
@@ -203,6 +203,72 @@ defmodule Req.UtilsTest do
              """
     end
 
+    test "it works with size" do
+      %{content_type: content_type, body: body, size: size} =
+        Req.Utils.encode_form_multipart([field1: {"value", size: 5}], boundary: "foo")
+
+      body = IO.iodata_to_binary(body)
+
+      assert size == byte_size(body)
+      assert content_type == "multipart/form-data; boundary=foo"
+
+      assert body == """
+             \r\n\
+             --foo\r\n\
+             content-disposition: form-data; name=\"field1\"\r\n\
+             \r\n\
+             value\r\n\
+             --foo--\r\n\
+             """
+    end
+
+    test "raise when size is not integer" do
+      assert_raise ArgumentError, fn ->
+        enum = Stream.cycle([1]) |> Stream.take(1)
+
+        Req.Utils.encode_form_multipart([field1: {enum, size: "not integer"}],
+          boundary: "foo"
+        )
+      end
+    end
+
+    test "can accept any enumerable" do
+      enum = Stream.cycle(["a"]) |> Stream.take(10)
+
+      %{body: body, size: size} =
+        Req.Utils.encode_form_multipart([field1: {enum, size: 10}], boundary: "foo")
+
+      body = body |> Enum.to_list() |> IO.iodata_to_binary()
+
+      assert size == byte_size(body)
+    end
+
+    test "blindly trust :content_length option" do
+      enum = Stream.cycle(["a"]) |> Stream.take(10)
+      advertised_length = 50
+
+      %{body: body, size: size} =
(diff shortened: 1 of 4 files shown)


knightpp:
See previous PR https://github.com/wojtekmach/req/pull/444

test/req/utils_test.exs
```diff
@@ -203,6 +203,62 @@ defmodule Req.UtilsTest do
...
+      body = IO.iodata_to_binary(body)
+
+      assert size == byte_size(body)
+      assert content_type == "multipart/form-data; boundary=foo"
+
+      assert body == """
+             \r\n\
+             --foo\r\n\
+             content-disposition: form-data; name=\"field1\"\r\n\
+             content-length: 5\r\n\
```
wojtekmach, line 229: Per https://datatracker.ietf.org/doc/html/rfc7578#section-4.8:

>   The multipart/form-data media type does not support any MIME header
>   fields in parts other than Content-Type, Content-Disposition, and (in
>   limited circumstances) Content-Transfer-Encoding.  Other header
>   fields MUST NOT be included and MUST be ignored.

Apologies for the confusion. We should call the option `:size` and its only purpose is to help calculating the content-length of the entire http request.

Req supports setting body to enumerable and will automatically (or rather Finch/Mint will) set the transfer-encoding to chunked. However, if the content-length is given, there's no need to do that:

```elixir
iex> Req.post!("https://reqbin.org/inspect", body: Stream.cycle(["abc"]) |> Stream.take(3)).body["headers"]
%{
  "accept-encoding" => ["zstd, br, gzip"],
  "host" => ["reqbin.org"],
  "transfer-encoding" => ["chunked"],
  "user-agent" => ["req/0.5.8"],
  "via" => ["1.1 fly.io"]
}
iex> Req.post!("https://reqbin.org/inspect", headers: [content_length: 9], body: Stream.cycle(["abc"]) |> Stream.take(3)).body["headers"]
%{
  "accept-encoding" => ["zstd, br, gzip"],
  "content-length" => ["9"],
  "host" => ["reqbin.org"],
  "user-agent" => ["req/0.5.8"],
  "via" => ["1.1 fly.io"]
}
```

WDYT?

wojtekmach: To add more context, there's no need to specify content-length in multipart _part_ because parts are delimited by the boundary (like `--foo--\r\n\`) anyway.

Again sorry for confusion from my end.

knightpp: Ahh, I see. So if we don't set content_length it will be treated as chunked encoding. Then do I need to change https://github.com/wojtekmach/req/blob/f5df0a54bc440f6ad4ff5837a54bde94acbeb102/lib/req/steps.ex#L434-L439 to set content-length only when there were no enum parts? 

Like this
```elixir
data = request.options[:form_multipart] ->
        multipart = Req.Utils.encode_form_multipart(data)

        m = %{request | body: multipart.body}
        |> Req.Request.put_new_header("content-type", multipart.content_type)
		
		if not multipart.contains_streaming_parts do # <<<
        	m |> Req.Request.put_new_header("content-length", Integer.to_string(multipart.size))
		else
			m
		end
```

wojtekmach: right, maybe it can even be something like:

```elixir
if size = multipart.size do
  Req.Request.put_new_header(req, "content-length", Integer.to_string(size)
else
  req
end
```

knightpp: What's changed
- `:size` is optional now
- when any part returns nil, the resulting size is nil
- when the resulting size is nil, no content-length header
- add two tests, check for nil and check for empty content-length header

wojtekmach:
I added some minor updates to the PR. Thank you!

PR #458, opened 2025-02-17
Link to HttpCookie

reisub:
I just released HttpCookie v0.7.0, with sane defaults for cookie length and count limits, and an easy to use Req plugin.


wojtekmach:
Thanks!

Issue #371, opened 2024-06-11
Would it make sense to make default retry functions public?


jozuas:
I'm thinking that it would be nice to have the ability to define your own `retry` function, and in the `retry` function if the conditions that I care about don't match, default to https://github.com/wojtekmach/req/blob/85bd7437488eeaf013659d1bebdce0511333d4ae/lib/req/steps.ex#L2292

I'll be happy to make a PR if this sounds reasonable.

wojtekmach:
This is very arbitrary but I'd like to keep the Steps to just have steps functions as it's already pretty big. The downside is we can't have this function which I agree would be very convenient to have. Perhaps there is another way to solve this?

jozuas:
I had a look at `Req.Steps` to see if there's a more general pattern that we could take advantage of, but sadly it seems that only the `transient?` function would benefit from being exposed. Creating a whole new module e.g. `Req.Conveniences` would not look so great if there's only one function there of interest.

The `retry` function implementation can return a tuple `{:delay, milliseconds}`, maybe we could support new options `{:retry, :transient}` and `{:retry, :safe_transient}`?

```elixir
    Req.new(
      ...,
      retry: fn
        _req, %Req.Response{status: 403} ->
          # Hypothetical service that returns HTTP 403 instead of 429 for "Too Many Requests" 
          {:delay, 30_000}

        _req, _resp ->
          {:retry, :safe_transient}
      end
    )
```

jozuas:
@wojtekmach how does the proposal sound? Is this something I could start working on that the Req library would be happy to accept?

wojtekmach:
Sorry I'm still considering it.

wojtekmach:
@andreyvolokitin can you share how you are using retry? You can set just the :retry_delay option to 0 in tests.

andreyvolokitin:
I can control retry delay and test backoff logic with these Req options, while still having Req to handle transient and `Retry-After` logic for me behind the scenes, so I no longer need any suggestions from this issue:

```elixir
    [
      retry: :transient,
      retry_delay: &Util.retry_backoff_exponential(&1, Application.fetch_env!(:app, :api_retry_start_delay_ms)),
      max_retries: Application.fetch_env!(:app, :api_max_retries)
    ]
```

[...]

Issue #296, opened 2024-01-31
`retry`: Allow changing of initial request parameters


Kraigie:
I think it would be useful to have some pathway to modify the initial request before retrying in cases where failure is expected and can be rectified. 

Say you make a request to an endpoint with an expired token. This would allow us to request a new token before retrying. It's totally possible that this functionality should live outside of the existing "step" functionality offered by this library, but figured I'd throw it up for discussion. It's also possible this functionality exists elsewhere in which case please let me know as I'm pretty new to the project!

I imagined it to look something like this

```elixir
Req.get("https://www.example.com/some/private/resource",
  auth: {:bearer, "foo"},
  retry: fn request, _response ->
    {true, Req.Request.put_header(request, "authorization", "Bearer bar")}
  end
)
```

Thoughts?

wojtekmach:
Thanks for bringing this up.

I'm curious about more use cases around this. If it's mostly around auth, I was thinking about something like this for a while anyway:

```elixir
auth: {:bearer, fn -> "foo" end}
```

would that help?

Kraigie:
It's not super clear to me how that's functionally different from the existing behavior unless that auth step you've proposed is run when encountering an error, taking in a function that returns a new token. If that's how I'm supposed to read it then I think that makes a lot of sense and would solve my issue! 

I'm struggling to think of a use case for my initial issue besides auth, so I think you're right that it should probably live in that space. Thinking on it, if we went with my initial suggestion we'd likely want to include what retry attempt we were on as well, and at that point I think it's pretty messy.

wojtekmach:
The difference would be every time we make a request (eg retry) we’d call a function to grab the auth as opposed to using the same value.

FWIW we can do that today:

```elixir
Req.new()
|> Req.Request.prepend_request_steps(custom_auth: fn r ->
  Req.update(r, auth: {:bearer, token})
  # or set the header directly here instead
end)
```

Thinking about it some more, I hope above is good enough. I’m gonna keep the issue open for a while in case some more ideas appear.

Regarding retry count, you can grab it from req.private. I plan to document it so it can be relied upon going forward.

Kraigie:
Thanks for the clarification!

Upon further review I think this is sufficient. Most APIs will return how long the token is good for and we can use that to determine if we need to fetch a new token before making a request. This should be able to be handled how you proposed!

[...]

Issue #202, opened 2023-07-13
Make retry interface a bit more flexible


DaTrader:
The `Req.Steps.retry/1` docs state that the `retry-after` header is only read in case of a 429 status, while the combo of the `retry` and `retry_delay` function options does not permit deciding on whether to retry say, based on the status, and then returning the delay retrieved from the `retry-after`.

Can this be made a little bit more flexible? I for instance need the functionality reserved for the 429 status in :retry_delay but for the 503.

Thanks

wojtekmach:
We could look at `retry-after` for 503s too. Could you send a PR?

DaTrader:
Sure, I can try over the weekend.

wojtekmach:
Closing in favour of #203.

DaTrader:
@wojtekmach 

In absence of a better place to ask you this, did you have issue reports of the following:

```
[error] Task #PID<0.2031.0> started from #PID<0.2030.0> terminating
** (Mint.TransportError) non-existing domain
    (req 0.3.11) lib/req.ex:958: Req.request!/2
```

This happens to me from time to time and I don't really know what to do with it. The only thing that helpers is shutting the LV instances with all of its processes.

Maybe related, I manage the timeouts myself by killing the async task (my async task that among other things such as caching, invokes Req) when the server is taking too long to respond. Can this be the cause of such behavior?

Also, it'd be good to know more on on the circumstances of this "non existing domain" error, for I change exactly nothing when I shut the LV instance down and it beings working again properly.

DaTrader:
The problem is when this happens, it does not stop (all subsequent requests raise until the LV instances are shut down).

DaTrader:
> If this error happens randomly ie the domain is actually good, search for :nxdomain on Mint and Finch issue tracker and see if something pops up.

Just to update you on this issue. I have replaced ALL calls of the bang functions (that raise the Mint exceptions) with the calls to functions that return errors and the thing now miraculously heals itself (after a series of logged Mint transportation errors that would otherwise raise) under one condition: one must not kill the process invoking Mint/Finch through Req, even when started as an async Task that is no longer needed (given the fact that it went into a caller managed timeout). Killing such task would mess up something with Mint/Finch internals.

The take away: IMO you should recommend using non-bang functions and letting async tasks complete on their own (the process using them must treat that explicitly if a GenServer, i.e. a LiveView instance or a scheduler or a cache process).

wojtekmach:
[...]

PR #464, opened 2025-03-05
retry_delay_in_ms: support float delays

ssaunderss:
When retry delays were introduced in https://github.com/wojtekmach/req/pull/226 it was implied that the retry delay would be greater than or equal to 1s. In situations where retry delays are either passed back as floats and/or less than 1s, it would be nice to use full precision since we're converting to milliseconds anyway.

This specifically fixes a problem in my case where an API is consistently handing back retries that are less than 1s, i.e. 0.314s.

### DIFF
--- a/lib/req/response.ex
+++ b/lib/req/response.ex
@@ -263,9 +263,11 @@ defmodule Req.Response do
   end
 
   defp retry_delay_in_ms(delay_value) do
-    case Integer.parse(delay_value) do
+    case Float.parse(delay_value) do
       {seconds, ""} ->
-        :timer.seconds(seconds)
+        seconds
+        |> :timer.seconds()
+        |> floor()
 
       :error ->
         delay_value


wojtekmach:
Thank you for the PR. Could you add a unit test?

wojtekmach:
Could you double check if fractions are allowed by RFC 9110? If not we shouldn’t change this parsing, perhaps the issue is somewhere else.

ssaunderss:
I think it's a bit of a grey area - RFC 9110 dictates that it should be a non-negative integer, but APIs / rate limiters often pass back fractional values because they're just doing a raw date diff. Maybe a better way to handle this is to continue parsing as an integer, but handle the case where a fractional part is present by just adding an extra second?

It would be nice to handle this at the library level instead of having to add boilerplate for handling errors downstream.

ssaunderss:
I made some updates based on the approach above and added some test coverage for `get_retry_after`

ssaunderss:
@wojtekmach I've made the suggested changes and added tests. Would you have time to review this approach of handling fractional delays by rounding up? Let me know if I should close this PR or if you'd like additional changes.

wojtekmach:
Apologies for delay. My bad. I think your solution is pragmatic however to be completely honest I'm not keen on accommodating broken servers that do not follow the spec. Looks like curl and Python requests ignore such invalid value for what it's worth.

We have some escape hatches in https://hexdocs.pm/req/Req.Steps.html#retry/1. Would any of these work for you? Using `retry: &fun/2` you could see if the header is there and if parsing with `Req.Response.get_retry_after/1` is unsuccessful, try parsing it as a float and return `{delay, ms}`.

If the current `retry: fun` interface would not work for you I'm much more keen on making it more useful to solve this and other edge cases. We have some open issues around this so if you'd be interested in going in that direction check them out first:

- #202
- #296
- #371

ssaunderss:
Thanks for the response. I believe I should be able to handle it using a custom retry function. In the meantime, I'll also take a look into those related issues!

PR #466, opened 2025-03-10
retry: Consider http2 :disconnected and :read_only as transient

DeltaEvo:
Those correspond to two Finch issues:
- https://github.com/sneako/finch/issues/216
- https://github.com/sneako/finch/issues/144

where retrying seems to be the only solution currently.


DeltaEvo:
I'm not fully convinced this is the "right" solution, but opening this to start a discussion if not

DeltaEvo:
Also didn't found tests for `:unprocessed`, I wasn't sure if I should add one as it feels like a change detector test, let me know if I should add some tests.

PR #468, opened 2025-03-15
Minor improvements to the documentation

sabiwara:
Hi @wojtekmach 👋

- fixes a small example to look more natural
- document the options from this [step](https://hexdocs.pm/req/Req.Steps.html#handle_http_errors/1) upstream

I also had a small question regarding [`default_options`](https://hexdocs.pm/req/Req.html#default_options/0): I noticed based on the implementation that it should be possible to set this from a config, but this is undocumented behavior so I didn't want to rely on it. Do you think this could this be officially documented?

(might avoid it anyway, I could totally understand the argument that such a global would be a smell, but given that the option exists and that it could streamline things in some non-library projects I figured I'd ask).

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -312,6 +312,16 @@ defmodule Req do
 
     * `:max_redirects` - the maximum number of redirects, defaults to `10`.
 
+  Other response options:
+
+    * `:http_errors` - how to handle HTTP 4xx/5xx error responses (via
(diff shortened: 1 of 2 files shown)


wojtekmach:
RE default_options, good call out. I plan to revamp Req.Test docs mentioning it. In my mind the only reasonable place to call it is test_helper.exs.

PR #469, opened 2025-03-20
Add links to documentation for Finch and Mint options

lawik:


lib/req.ex
```diff
@@ -383,12 +383,12 @@ defmodule Req do
 
         * A _module_ plug: a `module` name or a `{module, options}` tuple.
 
-  Finch options ([`run_finch`](`Req.Steps.run_finch/1`) step)
+  Finch options ([`run_finch`](`Req.Steps.run_finch/1`) step), [see Finch options](https://hexdocs.pm/finch/Finch.html#start_link/1):
```
wojtekmach, line 386: Since Req depends on Finch and Mint we can do something like the following, wdyt?

```markdown
see `Finch.start_link/1` for options
``` 

or


```markdown
[see Finch options](`Finch.start_link/1`)
```

lawik: Oh, cool :)

TIL

wojtekmach:
Thanks!

PR #475, opened 2025-05-24
Fix leading newline before form multipart body

pguillory:
Currently on form multipart requests, an unnecessary extra CR/LF gets inserted after the request headers. The fix is to change this encoding pattern:

    [crlf part] [crlf part] [crlf footer]

To this pattern:

    [part crlf] [part crlf] [footer]


wojtekmach:
Thank you for the PR. Could you fix tests?

PR #480, opened 2025-06-11
Automatically parse request body when using `:plug` option

Gazler:
To match automatically fetching the query_params, the request body is now also parsed when using the `:plug` option in the request. This is available on `conn.body_params`. The raw body is also stored as a binary, which can be read using the new `Req.Test.raw_body/1` function.

This simplifies some of the tests which were calling `Plug.Parsers` manually.


lib/req/test.ex
```diff
@@ -645,4 +645,48 @@ defmodule Req.Test do
...
+  ## Examples
+
+      iex> plug = fn conn ->
+      ...>   send(self(), {:raw_body, Req.Test.raw_body(conn)})
+      ...>   Plug.Conn.send_resp(conn, 200, "Hello, World!")
+      ...> end
+      iex>
+      iex> Req.post(plug: plug, json: %{hello: "world"})
+      iex> receive do {:raw_body, body} -> body end
+      "{\\"hello\\":\\"world\\"}"
```
wojtekmach, line 671:
```suggestion
      iex> echo = fn conn ->
      ...>   body = Req.Test.raw_body(conn)
      ...>   Plug.Conn.send_resp(conn, 200, body)
      ...> end
      iex>
      iex> resp = Req.post!(plug: echo, json: %{hello: "world"})
      iex> resp.body
      "{\\"hello\\":\\"world\\"}"
```

lib/req/test.ex
```diff
@@ -645,4 +645,48 @@ defmodule Req.Test do
...
+  @doc false
+  def read_request_body(conn, opts) do
+    with {:ok, body, conn} <- Plug.Conn.read_body(conn, opts) do
+      conn = update_in(conn.private[:raw_body], &((&1 || "") <> body))
+      {:ok, body, conn}
+    end
+  end
+
+  @doc """
+  Read the raw request body from a plug request.
```
wojtekmach, line 660:
```suggestion
  Reads the raw request body from a plug request.
```

lib/req/steps.ex
```diff
@@ -909,7 +909,10 @@ defmodule Req.Steps do
         * A _module_ plug: a `module` name or a `{module, options}` tuple.
 
       Req automatically calls `Plug.Conn.fetch_query_params/2` before your plug, so you can
-      get query params using `conn.query_params`.
+      get query params using `conn.query_params`. Req also automatically fetches the request
+      body using `Plug.Parsers` for JSON and multipart forms, you can get these using
+      `conn.body_params`. The raw request body of the request is avaialble by calling
```
wojtekmach, line 914:
```suggestion
      `conn.body_params`. The raw request body of the request is available by calling
```

lib/req/test.ex
```diff
@@ -645,4 +645,48 @@ defmodule Req.Test do
         """
     end
   end
+
+  ## Plug Parser
+
+  @doc false
+  def read_request_body(conn, opts) do
```
wojtekmach, line 652: people might be doing `import Req.Test`, wdyt about preventing this from being imported?

```suggestion
  def __read_request_body__(conn, opts) do
```

wojtekmach:
Thanks!

PR #481, opened 2025-06-14
add a request_retry_modifier function that can modify a request object before a retry

iwarshak:
**Summary**

This PR implements my proposal in https://github.com/wojtekmach/req/issues/296#issuecomment-2972003597.

It introduces a new retry option:

`{:delay, milliseconds, request_retry_modifier}`

The request_retry_modifier is a function that is invoked **before a retry**. It receives the current request and returns a modified request.

**Why?**

Currently, Req does not re-run steps on retries. This new option provides an escape hatch for scenarios where request modification is needed on retry—for example, recomputing a time-based authentication header.

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -354,6 +354,8 @@ defmodule Req do
 
             * `{:delay, milliseconds}` - retry with the given delay.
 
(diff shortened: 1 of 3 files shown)


wojtekmach:
Thanks for following up, this is great to hear. There is a bunch of retry related issues and PRs and I intend to do a full sweep on them eventually, apologies for the delay.

PR #487, opened 2025-07-05
Add some documentation about verify_on_exit!/1

sabiwara:


lib/req/test.ex
```diff
@@ -642,6 +644,11 @@ defmodule Req.Test do
   Sets a ExUnit callback to verify the expectations on exit.
 
   Similar to calling `verify!/0` at the end of your test.
+
+  This works as a ExUnit callback:
+
+      setup {Req.Test, :verify_on_exit!}
```
sabiwara, line 650: Note: I went with the qualified version because a) it doesn't imply an import b) it doesn't conflict with Mox'

But some other places like `set_req_test_from_context` use the atom version.

wojtekmach:
Thank you!

Issue #467, opened 2025-03-12
Allow to define what successful/unsuccessful response statuses are


jozuas:
I find that `Req` response handling requires a lot of boilerplate to handle due to the fact that all HTTP statuses are returned in the form of `{:ok, %Req.Response{status: some_http_status, ...}}`. While from the perspective of the library, `Req` has done its job, from the perspective of a user using `Req`, a non-200 HTTP status is often a failure case.
 
Example illustrating the point that all HTTP statuses are returned as `{:ok, ...}`:
```elixir
iex(tp@127.0.0.1)1> Req.post("https://httpbin.org/bearer")
{:ok,
 %Req.Response{
   status: 405,
   headers: %{
     "access-control-allow-credentials" => ["true"],
     "access-control-allow-origin" => ["*"],
     "allow" => ["GET, OPTIONS, HEAD"],
     "connection" => ["keep-alive"],
     "content-type" => ["text/html"],
     "date" => ["Wed, 12 Mar 2025 18:17:58 GMT"],
     "server" => ["gunicorn/19.9.0"]
   },
   body: "<!DOCTYPE HTML PUBLIC \"-//W3C//DTD HTML 3.2 Final//EN\">\n<title>405 Method Not Allowed</title>\n<h1>Method Not Allowed</h1>\n<p>The method is not allowed for the requested URL.</p>\n",
   trailers: %{},
   private: %{}
 }}
```

This means that when I write code, I generally have to have 3 branches of pattern matching for response handling:
```elixir
case Req.post("https://httpbin.org/bearer") do
  {:ok, %Req.Response{status: 200, body: body}} -> 
    # I do something with the body
    {:ok, result}

  {:ok, %Req.Response{status: status, body: body}} ->
    # I have to rewrap this response
    {:error, "Super useful error message or Exception struct"}

  {:error, exception} ->
    {:error, exception}
end
```

Now, if `Req` allowed to specify what the expected success status(es) are, this could end up being as simple as:
```elixir
with {:ok, %Req.Response{status: 200, body: body}} <- Req.post("https://httpbin.org/bearer", success_http_codes: [200]) do
  # I do something with the body
end
```

Assuming that `Req.post("https://httpbin.org/bearer", success_http_codes: [200])` would now return something like`{:error, %Req.HTTPError{status: 405, body: body}}`.

yordis:
[...]

Issue #193, opened 2023-06-20
`handle_http_errors`: Support `:response` and `:exception`


wojtekmach:
Extracted from https://github.com/wojtekmach/req/pull/188.

Example:

```elixir
Req.get("https://httpbin.org/status/400")
#=> {:ok, %Req.Response{}}

Req.get("https://httpbin.org/status/400", http_errors: :exception)
#=> {:error, %Req.Error{}}

Req.get!("https://httpbin.org/status/400", http_errors: :exception)
** (Req.Error)
```

Exact names still tbd.

yordis:
I prefer `:return` or `:raise`/`:exception` since `@spec` uses `no_return` (notice the `return` naming) so it is a bit more familiar.

wojtekmach:
As mentioned in https://github.com/wojtekmach/req/pull/188#issuecomment-1599356011 it is common to see things like:

```elixir
case f() do
  {:ok, resp} when resp.status in 200..299 ->
  {:ok, _} -> raise ...
  {:error, reason} -> raise ...
end
```

Maybe we also add `:expect` option to handle matching on 200..299. For example: `expect: 200`, `expect: [200, 201]`, `expect: 200..299`. Setting `:expect` (or `:expect_status` or however we'll end up calling this) will automatically raise on 4xx and 5xx responses.

wojtekmach:
Aside from the comment above, another idea for an API is:

```elixir
http_error: :as_response (default) | :as_exception
```

yordis:
I like both ideas but notice that you use `as_` for each value, so maybe do `http_error_as` instead.

---

In my brain that I can not let go is the `:response` thing! I wish Elixir finalize and come up with a `Result` (like Rust) naming everywhere! 😭 But whatever, maybe ChatGPT would be ideal here to move on and remove the analyzes paralysis (at least for me).

wojtekmach:
I still can't decide on the API. I'm gonna close this but if someone finds it in the future, please feel free to continue discussing this.

It's not perfect but I think what we have today is sensible enough:

```elixir
Req.get!(req)

# if 4xx/5xx errors that would _not_ be automatically retried, do:
Req.get!(req, http_errors: :raise)

# if other 2xx/3xx might happen, do:
resp = Req.get!(req)
%{status: 200} = resp
```

yordis:
> Im happy to remove :raise because when people use bang functions we are raising anyway.

[...]

PR #188, opened 2023-06-14
feat: add ok property to response and to_tuple func

yordis:

PR #493, opened 2025-07-21
New step: `expect`

wojtekmach:
This is a replacement for https://hexdocs.pm/req/Req.Steps.html#handle_http_errors/1 which will be deprecated and removed in Req 1.0.

Some more context here:

- https://github.com/wojtekmach/req/pull/188
- https://github.com/wojtekmach/req/issues/193
- https://github.com/wojtekmach/req/issues/467

In a nutshell:

```diff
- case Req.request(req) do
-   {:ok, %{status: 200}} = result ->
-     result
-
-   {:ok, resp} ->
-     exception =
-       RuntimeError.exception("""
-       unexpected status #{resp.status}
-
-       #{inspect(resp.body, pretty: true)}
-       """)
-
-     {:error, exception}
-
-   {:error, _} = error ->
-     error
- end
+ Req.request(req, expect: 200)
```


wojtekmach:
Maybe we put it after decompress but before decode so the exception message just prints the string

yordis:
I still feel `ok` literally how Fetch Web API works is a good idea, along with `expect` thingy, and familiar to other ecosystems. 

`expect` still extremely valuable to control the flow from the domain perspective, while `ok` is useful in the infra (steps) perspective for most use cases.

At least for most OpenAPI, non-redirecting APIs; which are the most commons anyway.

I think it is worth it to keep it around, and trivial enough where it doesn't add much complexity, if any.

wojtekmach:
Thanks for the feedback about expect. I concede response.ok is ubiquitous due to being part of the Web platform.

I think in pattern matches, by the way of falling into another code branch that is for transport errors, expect "wins":

```diff
  case resp do
-   {:ok, %{ok?: true} = resp -> ...
-   {:ok, resp} -> ...
-   {:error, e} -> ...
+   {:ok, resp} ->
+   {:error, e} -> ...
  end
```

In other cases

```diff
- if resp.status in 200..299 do
+ if resp.ok? do
```

yeah, that looks really clean. My personal opinion though is it doesn't justify adding a new field to the struct even if that's some false scarcity argument.

yordis:
`ok` will always be 200..299 as the specs says, it is up the implementors to understand that behaviour. To be honest, the reason I keep coming back to this one (and I rather aligned here) is that it is such common case due to the popularity of JS and fetch.

Most likely useful for default steps behaviour and things like that, purely a semantic meaning that sometimes it is useful.

I have come across that `200..299` far too many times.

lib/req/steps.ex
```diff
@@ -2333,6 +2335,92 @@ defmodule Req.Steps do
...
+  > Since response headers/body can contain sensitive data, be careful about raising
+  > this error and automatically logging it, sending to exception trackers, etc.
+
+  ## Examples
+
+      iex> resp = Req.get!("https://httpbin.org/status/200", expect: 200)
+      iex> resp.status
+      200
+
+      iex> Req.get!("https://httpbin.org/status/404", expect: 200..299)
```
yordis:
```suggestion
      iex> Req.get!("https://httpbin.org/status/404", expect: :ok)
```

could be an atom representing a given config as well? 😅

wojtekmach:
anyone has any objections to calling the step and option a more explicit:

```elixir
expect_status: 200..299
```

?

I think it plays nicer which how I ended up calling the exception: Req.UnexpectedResponseStatusError. Could see shorter Req.UnexpectedStausError too, btw.

The main driver is there’s an `expect` header—-https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Expect

so :expect_status would totally disambiguate even if slightly more verbose.

thoughts?

PR #497, opened 2025-07-24
Include Private Data in Plug Connection

aj-foster:
Hello 👋🏼 

This is a proposal / for discussion purposes, but it was easier to demonstrate the proposal in code rather than as an issue.

Would you consider forwarding some of the original request data to the `conn` struct created by `run_plug`? Doing so could enable advanced testing workflows by client libraries that utilize Req. Perhaps there is other, similar data from the original request that is worth including as well.

Thanks for your consideration! ❤️


aj-foster:
The idea came from a client library where I'd like to implement some helpful wrappers around `Req.Test` — for example, being able to assert that a certain call was made based on the name of the client function (as opposed to the method/path/etc.). Information about the original call already fits nicely in the `Req.Request` private map, and it would be convenient to access it on the `conn` as well.

PR #498, opened 2025-08-02
Fix AWS SigV4 compatibility with Supabase Storage S3

meeq:
The [AWS SigV4 documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html) states:
> The CanonicalHeaders list must include the following:
> * HTTP host header.
> * If the Content-MD5 header is present in the request, you must add it to the CanonicalHeaders list.
> * Any x-amz-* headers that you plan to include in your request must also be added. For example, if you are using temporary security credentials, you need to include x-amz-security-token in your request. You must add this header in the list of CanonicalHeaders.

[Supabase Storage explicitly forbids certain headers from being included in the signature](https://github.com/supabase/storage/blob/681287a6c629b2a899516350e583cad13b33ed92/src/storage/protocols/s3/signature-v4.ts#L64-L79), so this patch pares down the canonical headers to a more broadly-supported subset. 

For reference, `botocore` (the underlying library that powers the canonical `aws-cli` tool) [only signs "interesting" headers, and omits the rest](https://github.com/boto/botocore/blob/ac5a0db86894822ba068762a2b199e2b22745b40/botocore/auth.py#L910-L940), so this change should not break compatibility.

Related to https://github.com/wojtekmach/req_s3/issues/22


wojtekmach:
Thank you, sorry about the delay. There were conflicts and I ended up using slightly different solution to this problem, let me know if you have any concerns!

PR #499, opened 2025-08-07
fix: set :path_params_template on empty params

davydog187:
This is to address https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/535

When you have a static path `/a/b/c`, the `:path_params_template`  is not set, even if you pass `path_params: []`. This PR disambiguates no path params with empty path params so that this can be checked by `OpenTelemetry.Req`

If this approach looks ok to you @wojtekmach I can add a test

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -495,7 +495,7 @@ defmodule Req.Steps do
 
   ## Request Options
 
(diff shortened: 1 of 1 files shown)


wojtekmach:
Sounds good!

PR #503, opened 2025-08-18
Use built in zstd for compression when provided

mattmatters:
For >=v28 of Erlang, Zstd is part of Erlang's stdlib. This greatly simplifies dependency requirements in Req.

I took some great pains to match how ezstd is called and matching its return type. It did prove to get a little ugly though to prevent compiler type warnings however.

Hopefully in several years this can be built with the expection zstd always exists.

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -306,7 +306,7 @@ defmodule Req.Steps do
 
     * `br` (if [brotli] is installed)
 
-    * `zstd` (if [ezstd] is installed)
+    * `zstd` (if your Erlang version is >= 28 or [ezstd] is installed)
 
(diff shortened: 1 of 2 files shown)


wojtekmach:
Thank you. The PR looks good to me. I need some more time to figure out if it's worth it to keep support for `ezstd` since it will be less need of that going forward.

wojtekmach:
Thank you for looking into it and apologies for delay. I'm gonna drop ezstd on main shortly!

mattmatters:
> Thank you for looking into it and apologies for delay. I'm gonna drop ezstd on main shortly!

No worries :) Thanks for building and maintaining an awesome library!

Issue #19, opened 2021-07-07
`auth`: Support digest auth


wojtekmach:
curl example:

```bash
$ curl --fail --user foo:bar --digest https://httpbin.org/digest-auth/auth/foo/bar
{
  "authenticated": true,
  "user": "foo"
}
```

We should support:

```elixir
Req.get!("https://httpbin.org/digest-auth/auth/foo/bar", auth: {:digest, "foo", "bar"})
```

wojtekmach:
Re A) yup, option 1, except it would be a response step. An error step would be when we didn’t get a response but a socket error and such.

Re B) I think it is actually a good enough reason not to support it. What I mean is, if someone needs it and is willing to do the work then I’d accept a PR but otherwise it is not a high priority. I wanted to have it for completeness but you are totally right it is mostly a thing of the past.

thanks!

zachallaun:
Sounds good! Agreed that it’s not a priority. It may be something I look to implement anyways as a small exercise. =)

PR #511, opened 2025-10-24
Support digest auth

regan-karlewicz:
Addresses #19 

I recently had a use case for digest auth and ended up implementing a Req plugin for it, so I decided to open a PR to integrate into Req after seeing the closed issue :)

### DIFF
--- a/lib/req/steps.ex
+++ b/lib/req/steps.ex
@@ -2070,12 +2070,17 @@ defmodule Req.Steps do
   end
 
   @doc """
-  Handles HTTP Digest authentication when auth is provided as `{:digest, userinfo}`
+  Handles HTTP Digest authentication.
+
+  This step is invoked when setting `:auth` option with `{:digest, ...}`. When response is HTTP 401 with `www-authenticate` header, this step will calculate `authorization: Digest ...` header and make another request.
+
+  See `auth/1`.
 
   ## Examples
 
-      iex> Req.get!("https://httpbin.org/digest-auth/auth/user/pass", auth: {:digest, "user:pass"})
-      {:ok, %Req.Response{status: 200, body: %{"authenticated" => true, "user" => "user"}}}
+      iex> resp = Req.get!("https://httpbin.org/digest-auth/auth/user/pass", auth: {:digest, "user:pass"})
+      iex> resp.status
+      200
   """
   @doc step: :response
   def handle_http_digest({request, %Req.Response{status: 401} = response}) do


davydog187:
Hi @wojtekmach 👋 this came up at TV Labs and I encouraged @regan-karlewicz to open a PR. Let us know if you want to chat!

lib/req/steps.ex
```diff
@@ -2059,6 +2079,220 @@ defmodule Req.Steps do
...
+    16
+    |> :crypto.strong_rand_bytes()
+    |> Base.encode16(case: :lower)
+  end
+
+  def hash_function(algorithm) do
+    case String.upcase(algorithm) do
+      "MD5" <> _ -> {:ok, &md5_hash/1}
+      "SHA-256" <> _ -> {:ok, &sha256_hash/1}
+      # TODO: Add support for SHA-512-256
```
wojtekmach: do you intend to implement this in this PR? If not, which is totally fine, let's remove this todo and wait until it comes up.

lib/req/steps.ex
```diff
@@ -2059,6 +2079,220 @@ defmodule Req.Steps do
...
+      _ -> {:error, :unsupported_digest_algorithm}
+    end
+  end
+
+  defp build_auth_header(opts) do
+    header_parts =
+      opts
+      |> Keyword.take([:username, :realm, :nonce, :uri, :response, :opaque])
```
wojtekmach: Let's use Keyword.validate!/2 to make sure people didn't accidentally typo.

wojtekmach (commented):
Thank you for the PR! I left some comments below.

lib/req/steps.ex
```diff
@@ -2059,6 +2069,232 @@ defmodule Req.Steps do
...
+        {request, response}
+    end
+  end
+
+  def handle_http_digest(other), do: other
+
+  defp handle_challenge_reply(request, response, opts) do
+    username = Keyword.get(opts, :username, "")
+    password = Keyword.get(opts, :password, "")
+    count = Keyword.get(opts, :count, 1)
```
wojtekmach: `:count` seems to be always 1? The only call site I found is:

```elixir
        handle_challenge_reply(request, response,
          username: username,
          password: password
        )
```

which does not set :count.

wojtekmach: Ah, I suppose the response for the second request would not have www-authenticate header anymore so we'd have to store it for later and add API to grab it, perhaps something like `Req.Response.get_auth_digest_data(resp)` but yeah, not loving it 😅

regan-karlewicz: Yes, your understanding is correct about the use case here (reducing # of requests)

I do like the idea of a helper function to retrieve the data and passing it back in on subsequent requests, since it gives the user full control and doesn't require maintaining an internal state. It could retrieve the data needed for subsequent requests in regard to the current nonce and count (depending on whether a `nextnonce` was returned)

I agree, let's wait for now if that sounds good to you, if we need this later I can take a shot at implementing it :)

lib/req/steps.ex
```diff
@@ -2059,6 +2069,232 @@ defmodule Req.Steps do
     {request, response}
   end
 
+  @doc """
+  Handles HTTP Digest authentication when auth is provided as `{:digest, userinfo}`
+
+  ## Examples
+
+      iex> Req.get!("https://httpbin.org/digest-auth/auth/user/pass", auth: {:digest, "user:pass"})
+      {:ok, %Req.Response{status: 200, body: %{"authenticated" => true, "user" => "user"}}}
```
wojtekmach:
```suggestion
  Handles HTTP Digest authentication.

  This step is invoked when setting `:auth` option with `{:digest, ...}`. When response is HTTP 401 with `www-authenticate` header, this step will calculate `authorization: Digest ...` header and make another request.

  See `auth/1`.

  ## Examples

      iex> resp = Req.get!("https://httpbin.org/digest-auth/auth/user/pass", auth: {:digest, "user:pass"})
      iex> resp.status
      200
```

wojtekmach:
Thank you!

PR #514, opened 2025-10-29
fix: tiny typo

davydog187:
was looking at the diff since the last release, and noticed that I had a tiny typo in #499


wojtekmach:
Thanks!

PR #535, opened 2026-03-27
Honor configured :retry_delay over Retry-After

### LINKED ISSUE #463: `:retry_after` ignored if there is a Retry-After response header
The documentation for `:retry_delay` says:

> if not set, which is the default, the retry delay is determined by the value of the Retry-After header on HTTP 429/503 responses. If the header is not set, the default delay follows a simple exponential backoff: 1s, 2s, 4s, 8s, ...

This makes it seem like the priority would be:
- Use the configured `:retry_delay` option
- If not provided, use the Retry-After header
- If there's no header, use the backoff strategy

But it seems that it's really:
- If the Retry-After header is present, use it
- If there's no header, use the `:retry_delay` option
- If that's not set, use the backoff strategy

https://github.com/wojtekmach/req/blob/ba4de2a7deb0085a74617faf3ee099074ce9b541/lib/req/steps.ex#L2210-L2238

I'm unsure if this is a bug and it should always use the `:retry_after` option if it's set, or if it's a documentation fix to clarify that `:retry_after` is ignored when there's a Retry-After header.

Sim-hu:
Fixes #463.

The docs say `:retry_delay` should take precedence when it is explicitly set,
but 429/503 responses were still using `Retry-After` first.

This changes the retry delay lookup so that an explicit `:retry_delay` wins,
and `Retry-After` is only used when `:retry_delay` is not set.

Also adds a regression test for the case where both are present.


wojtekmach:
Thank you!

PR #538, opened 2026-04-14
fix: allow :finch option with IPv6 URLs

meox:
Req auto-sets `inet6: true` when the URL host contains `:` (e.g.
`http://[::1]:8080/`). The `finch_name/1` guard treated `:inet6` as a
custom pool option, causing it to raise "cannot set both :finch and
:connect_options" when a named Finch pool was also provided.

Narrow the guard to only reject `:connect_options`, which would
genuinely be ignored by a pre-configured Finch instance. The `:inet6`
flag is orthogonal to pool selection and should not conflict.

This fixes connection pool isolation for services that use separate
named Finch pools when the target is an IPv6 address.

### DIFF
--- a/test/req/finch_test.exs
+++ b/test/req/finch_test.exs
@@ -214,6 +214,45 @@ defmodule Req.FinchTest do
       end
     end
 
+    test ":finch and :inet6" do
+      # :inet6 can be auto-set for IPv6 URLs; it should not conflict with :finch
+      assert_raise ArgumentError, "unknown registry: MyFinch", fn ->
+        Req.get!("http://localhost", finch: MyFinch, inet6: true)
+      end
+    end
+
+    test ":finch with IPv6 URL" do
+      start_supervised!(
+        {Plug.Cowboy,
+         scheme: :http,
+         plug: ExamplePlug,
+         ref: ExamplePlug.IPv6Named,
+         port: 0,
+         net: :inet6,
+         ipv6_v6only: true}
+      )
+
+      ipv6_port = :ranch.get_port(ExamplePlug.IPv6Named)
+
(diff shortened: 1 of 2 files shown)


test/req/finch_test.exs
```diff
@@ -214,6 +214,45 @@ defmodule Req.FinchTest do
...
+         pools: %{
+           default: [
+             protocols: [:http1],
+             conn_opts: [transport_opts: [inet6: true]]
+           ]
+         }}
+      )
+
+      # Before the fix, this would raise "cannot set both :finch and :connect_options"
+      # because Req auto-sets inet6: true for IPv6 URLs like [::1]
```
wojtekmach:
```suggestion
```

test/req/finch_test.exs
```diff
@@ -214,6 +214,45 @@ defmodule Req.FinchTest do
       end
     end
 
+    test ":finch and :inet6" do
+      # :inet6 can be auto-set for IPv6 URLs; it should not conflict with :finch
+      assert_raise ArgumentError, "unknown registry: MyFinch", fn ->
+        Req.get!("http://localhost", finch: MyFinch, inet6: true)
+      end
```
wojtekmach: the intention is we do NOT fail with `"cannot set both :finch and :connect_options"`, right?

If so, why don't we move this assertion as the last one in the test, perhaps something like this?

```elixir
assert Req.request!("http://localhost:#{port}", finch: finch_name, inet6: true).body == "ok"
```

wojtekmach (commented):
Good catch, thank you!

wojtekmach:
Thank you!

PR #540, opened 2026-04-25
Warn and strip userinfo if present in url

sabiwara:
Currently, when an URL with a `userinfo` like `Req.get!("https://foo:bar@httpbin.org/basic-auth/foo/bar")` is provided, `Req` silently ignores the userinfo, which makes it non-obvious why the request returns a 401:

It is especially confusing since `req.url` is still showed with the userinfo, and it can take some time to figure out when migrating an existing project from a library that used to handle the userinfo.

This PR adds a warning to shorten the feedback loop and provide actionable steps to address the issue.
We could probably be raising here, I went with a warning to be more conservative and kept the existing behavior - happy to update the strategy if we want to handle it differently.

Screenshot 2026-04-25 at 17 03 51

### DIFF
--- a/lib/req.ex
+++ b/lib/req.ex
@@ -149,6 +149,7 @@ defmodule Req do
   > For interopability with those, use
   > `Req.get_headers_list/1`.
   """
+  require Logger
 
   # Response streaming to caller:
   #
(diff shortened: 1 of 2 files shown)


jswanner:
While I agree that Req shouldn't silently ignore `userinfo` in a URL, but since Req is "batteries-included" I'd more expect it to "just work" instead of warn, IMHO.

wojtekmach:
Closing in favour of https://github.com/wojtekmach/req/pull/541

PR #542, opened 2026-05-01
feat(run_finch): Allow setting `conn_max_idle_time` via Req interface

Slotos:
Additionally, this patches a case where presence of `pool_max_idle_time` alone wouldn't trigger custom pool spawn, effectively ignoring the setting.

Frankly speaking, this is a quickfix. IMO a proper fix should expose `adapter_options` and, for Finch specifically, validate whether it's a keyword list with no content enforcement. This would eliminate the abstraction leak entirely by narrowing the scope of responsibility.

I'm proposing this quickfix as a way to apply fix for AWS Nat Gateway connection termination issue discussed in
https://github.com/sneako/finch/issues/272. While `:pool_max_idle_time` can help, it doesn't address the issue as directly as `:conn_max_idle_time` does.


wojtekmach:
Thank you for the PR and sorry about the delay.

I've been thinking about something along the lines of `adapter_options` for a while.

So as an aside, in theory Req supports multiple adapters but we obviously just ship with Finch adapter. Some of the options, including :conn_max_idle_time, are finch specific so when switching to another adapter, that option would probably be silently ignored. A lot of options like :inet6 or :cacerts should be supported by every adapter, however.

We recently extended `:finch` option to be:

```elixir
finch:
  Finch.name() |
    {Finch.name(), options} # currently only `pool_tag: tag`
```

And so I think a more extensible mechanism could be:

```elixir
finch:
  Finch.name() |
    {Finch.name(), [Finch.request_opt() | Finch.Request.build_opt()]} |
    [Finch.request_opt() | Finch.Request.build_opt() | pool_opt()]
```

For example:

```elixir
Req.get!(url, finch: [conn_max_idle_time: 10000])
Req.get!(url, finch: {MyFinch, pool_timeout: 5000})
```

And then we would deprecate Finch-specific options in favour of passing them this way.

The only caveat is we won't be able to pass options as is as they need to land in one of three buckets:

```elixir
Finch.build(method, url, headers, body, opts)
                                        ^^^^

Finch.request(req, name, opts)
                         ^^^^

Finch.start_link(pools: %{default: opts})
                                   ^^^^
```

so we'd have to keep options mapping in sync with Finch but I think that's an ok tradeoff.

This is obviously out of the scope of this PR. I'm curious if this `finch: options` design resonates with you at all?

meox:
I suggest to make it uniform:

```elixir
Req.get!(url, finch: [pool: MyFinch, conn_max_idle_time: 10000])
```

removing the tuple notation and add a new `pool` or `name` field where the user can specify the name

wojtekmach:
Thanks @meox, maybe :name indeed since it would match Finch.start_link(name: MyFinch)

`Req.get!(url, finch: [name: MyFinch])`

jswanner:
Yeah, I agree with @meox that simplifying the shape of the finch-specific options to a keyword list is best, and I also think going with `:name` for that option name is best.

Something I was just thinking through -- as Req shipping with multiple adapters is a little less theoretical 😉, and if that were to land -- how one would specify which builtin adapter they wanted to use and provide adapter specific options to it. But, instead of complicating the `:adapter` option to allow for adapter-specific options there (`adapter: {:httpc, [some_httpc_option: value]}`), I think something like `Req.get!(url, adapter: :httpc, httpc: [some_httpc_option: value])` is better anyway.

wojtekmach:
Yeah my plan is to have `finch:`, `httpc:`, etc, and I consider warning if you set :finch options but ended up using httpc thus making those other options ignored.

PR #548, opened 2026-06-18
Allow finch 0.23.0

### LINKED ISSUE #375: `:pool_not_available` returned during pool startup window introduced in 0.22.0
(This issue was written by Claude after some extensive troubleshooting trying to understand why there was suddenly `:pool_not_available` errors surfacing directly after every deploy. I have limited understanding of Finch's inner workings, so posting this in the hope that it is at least pointing in the right direction. The issue is certainly real, it started showing up immediately after upgrade 0.2.1 -> 0.2.2)

In Finch 0.22.0, the new per-pool supervisor + registry-backed tracking introduces an observable race window between pool *supervisor* registration and pool *worker* registration. A second request arriving during that window receives `Finch.Error{reason: :pool_not_available}` even though the first request triggered pool creation milliseconds earlier and will succeed.

This is observable in production immediately after every node restart, when a burst of parallel outbound requests fires before any per-host pool has fully come up. We see it consistently after every deploy on multiple integrations after upgrading from 0.21.0 to 0.22.0.

### Reproduction (conceptual)

1. Finch instance with a default pool config (`protocols: [:http1]`).
2. No prior request to `api.example.com`, so no pool exists.
3. Fire N concurrent requests to `api.example.com`. (N ≥ 2 is enough on a busy scheduler.)
4. Request A wins the race into `Pool.Manager.maybe_start_pool/4` and calls `start_pool/3`.
5. Request B arrives between (a) Finch.Pool.Supervisor being registered in `supervisor_registry_name` via `{:via, Registry, ...}` — which happens at process-spawn time, before `Supervisor.init/1` even runs — and (b) the supervisor finishing startup of its NimblePool children (which register themselves in `registry_name` during `init_worker`).
6. Request B's `do_get_pool/4` finds `Registry.lookup(registry_name, pool_name) == []` and falls into `maybe_start_pool/4`. The supervisor lookup succeeds (entry from step 5a), so the function returns `:not_ready`.
7. `Finch.__stream__/6` maps `:not_ready` (and `:not_found`) into `Finch.Error.exception(:pool_not_available)`.
8. Caller sees a transient failure for what should have been a benign concurrent first-touch.

Relevant code:

- `lib/finch/pool/manager.ex` lines 98-114, 122-135
- `lib/finch/pool/supervisor.ex` lines 10-12 (name is `{:via, Registry, ...}`, registers at spawn)
- `lib/finch.ex` around line 838 (`{:error, Finch.Error.exception(:pool_not_available), acc}`)

For HTTP/2 this window is widened further by the explicit "register only after `:connected`" change (#356), but **the same window exists for HTTP/1** because the supervisor is registered before child init.

### Why it's a regression from 0.21.0

In 0.21.0 the pool was effectively reachable synchronously after `start_pool/3` returned — there was no separate "supervisor known but workers not yet ready" state machine to fall through. The new architecture in #344 introduces this state, but the manager surfaces it to callers as a fatal `:pool_not_available` instead of, say, waiting briefly for the workers to register or retrying internally.

### Workarounds we've applied

- Custom Req retry predicate that catches `Req.HTTPError{protocol: :http2, reason: :pool_not_available}` and retries. (Side note: Req mislabels every `Finch.Error` as `protocol: :http2` regardless of the actual protocol — separate issue for the Req side.)
- Considered pre-warming pools at boot.

Neither is satisfying — the surface a transient race shouldn't propagate to callers.

### Suggested fixes (any one of)

- In `Pool.Manager.maybe_start_pool/4`, if the supervisor is present but workers aren't, **block briefly** (configurable `:pool_ready_timeout` or similar) waiting for the worker registration, then re-lookup. The pool supervisor already knows when its children have started — surface that via a `:syn`-style synchronization or a small `:gen.call` to the supervisor.
- Alternatively, register the worker key in `registry_name` *eagerly* in `Finch.Pool.Supervisor.init/1` (or just before children start) so the `:not_ready` branch never fires for HTTP/1, and limit `:not_ready` to the HTTP/2 not-yet-`:connected` case that's intrinsically async.
- At minimum, document the new behavior loudly in the upgrade notes and recommend `retry: :transient` callers extend their predicate to include `:pool_not_available`.

### Environment

- Elixir 1.19.5, OTP 27.x
- Finch 0.22.0 (upgraded from 0.21.0)
- Req 0.5.17 (default `Req.Finch` instance, `protocols: [:http1]`)
- Mint 1.8.0
- All HTTP/1 traffic — but the error is labeled `protocol: :http2` because of Req's `Finch.Error` normalization.

### Related

- #344 (the architectural refactor that introduced the per-pool supervisor + registry)
- #356 (the HTTP/2 "register only when connected" change, which widens the same window for HTTP/2)

sabiwara:
Finch 0.23.0 has been released and fixes https://github.com/sneako/finch/issues/375, but `Req` currently doesn't allow it.

I'm not sure that the approach scales, perhaps we should just go back to `"~> 0.21"`.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -90,7 +90,7 @@ defmodule Req.MixProject do
 
   defp deps do
     [
-      {:finch, "~> 0.21.0 or ~> 0.22.0", finch_opts()},
+      {:finch, "~> 0.21.0 or ~> 0.22.0 or ~> 0.23.0", finch_opts()},
       {:mime, "~> 2.0.6 or ~> 2.1"},
       {:jason, "~> 1.0"},
       {:nimble_csv, "~> 1.0", optional: true},


wojtekmach:
Thank you!

wojtekmach:
I don't remember rationale for being stricter about versions (aside from the usual which is 0.x) so yeah, I think either is OK!

ftes:
This hit us in a dependabot PR just now:

finch was updated and req was silently downgraded from `v0.6.1` to `v0.5.17` (which was probably the last version with the looser version range for finch).

Sounds like it would really be better to switch back to `~> 0.21`.

wojtekmach:
Req v0.6.2 allowing Finch ~> 0.21 is out.

PR #549, opened 2026-06-18
Run mix format, fix lint in CI

sabiwara:
`lint` CI was failing https://github.com/wojtekmach/req/actions/runs/27795214585/job/82253370092?pr=548.

Seems tests are failing too though.


wojtekmach:
Thank you!

Issue #556, opened 2026-08-06
No support for query parameters with the same key


mruoss:
Although this is not common, passing the same query parameter mutliple times is still expected by some servers.

Consider the following URL: https://httpbin.io/anything?a=1&a=2 and how we would use Req to construct it:
```ex
req = Req.new(url: "https://httpbin.io/anything", params: [a: 1, a: 2])
Req.get!(req).body["args"]
# expected: %{"a" => ["1", "2"]}
# actual: %{"a" => "2"}
```

Culprit: somewhere https://github.com/wojtekmach/req/blob/v0.7.2/lib/req/steps.ex#L623

wojtekmach:
Hello, thanks for the report. This was intentional, see https://github.com/wojtekmach/req/commit/d14eb4f0e3cd6404da5bef2d11dbb6c7de162973. You'd have to manually set query params, e.g.:

```elixir
iex> req = Req.new("https://httpbin.io/anything")
iex> req = update_in(req.url, &URI.append_query(&1, "a=1&a=2"))
iex> Req.request!(req).body["args"]
%{"a" => ["1", "2"]}
```

I'm going to close this but please feel free to continue discussing this.

wojtekmach:
If it's a significant downgrade for kubereq users, please lmk. Otherwise if it's not too much of a hassle, I'd prefer to not change it again!

mruoss:
No I think I can work around it. Thanks for the heads-up.

mruoss:
https://github.com/mruoss/kubereq/pull/194/changes#diff-d5076506e5cc95b8453c90eb47671444dce56e8ed79ce5503e22b9a6e47c87ca

jswanner:
In case anyone else ends up here... The approach I took in my application wasn't much different from what @mruoss linked to above, but to put the diff here:

```diff
- Req.request(params: params, url: "/path)
+ Req.request(url: URI.parse("/path") |> URI.append_query(URI.encode_query(params)))

**Comment by wojtekmach** [comment]
Thanks @jswanner. Maybe there's a middle ground where we always overwrite, which solves the original issue, but can overwrite with duplicates? I haven't tried that out but I'm thinking:

```elixir
request("?", params: [a: 1, a: 2])    #=> "?a=1&a=2"
request("?a=1", params: [a: 1, a: 2]) #=> "?a=1&a=2"
```

PR #561, opened 2026-08-21
Allow duplicate params while always overwriting existing

jswanner:
@wojtekmach I believe this implements the middle ground you were referencing in https://github.com/wojtekmach/req/issues/556#issuecomment-5338778008.


wojtekmach:
Thanks @jswanner, I used another approach on main.

PR #563, opened 2026-08-27
Fix retry-after header delay docs from seconds to milliseconds

vbalazs:
Updated documentation to reflect that the `retry-after` header delay value is now in milliseconds instead of seconds.


wojtekmach:
Thank you!

